PR #10, opened 2020-06-19
adds calculator unit tests

edumorlom:


covid19-dashboard/dc-covid19/src/DataCalculator.ts
```diff
@@ -1,90 +1,126 @@
...
+): DataPerGeoIdPerDate {
+    // If there isn't any data, return an empty {}
+    if (!Object.keys(data).length) {
+        console.log("No input data.")
+        return {}
+    }
 
-    // If there is data,
-    let outputData: DataPerGeoIdPerDate = {}
+    let date0 = moment(range[0]);
```
shifucun, line 39: const date0

covid19-dashboard/dc-covid19/src/DataCalculator.ts
```diff
@@ -1,90 +1,126 @@
...
 
 /**
  * Performs a calculationType iteratively on two dates.
  * @param data: input data in the form of dates->geoIds->value.
  * @param range: the range of dates to calculate the data for in ISO Format. Example: ["2020-01-01", "2020-01-10"]
- * @param deltaDays: Do we wanna compare the data every 2 days? every 7? Any number works.
  * @param calculationType: What type of calculation are we performing?
- * @param geoIdToPopulation: geoId->Population. After calculating the difference, we want to divide by total population
+ * @param deltaDays?: Do we wanna compare the data every 2 days? every 7? Any number works.
+ * @param geoIdToPopulation?: geoId->Population. After calculating the difference, we want to divide by total population
```
shifucun, line 25: I suggest not to do per-capita in this function. Instead have a separate function to do point-wise data conversion

covid19-dashboard/dc-covid19/src/DataCalculator.ts
```diff
@@ -1,90 +1,126 @@
...
+        // Add one day to the iterativeDate. Example: from "2020-01-01" to "2020-01-02".
+        date0 = date0.add(1, "days")
     }
     return outputData
+}
+
+/**
+ * Given two values (or one if absolute or absolutePerCapita), return the result of calculationType.
+ * @param values: [number, number] -> values[0] is date - deltaDays, values[1] is the actual date.
+ * @param population: some calculations require the population.
```
shifucun, line 95: also take the population out from here. So this function do binary operations on values

covid19-dashboard/dc-covid19/src/DataCalculator.ts
```diff
@@ -1,90 +1,126 @@
...
  * @param data: input data in the form of dates->geoIds->value.
  * @param range: the range of dates to calculate the data for in ISO Format. Example: ["2020-01-01", "2020-01-10"]
- * @param deltaDays: Do we wanna compare the data every 2 days? every 7? Any number works.
  * @param calculationType: What type of calculation are we performing?
- * @param geoIdToPopulation: geoId->Population. After calculating the difference, we want to divide by total population
+ * @param deltaDays?: Do we wanna compare the data every 2 days? every 7? Any number works.
+ * @param geoIdToPopulation?: geoId->Population. After calculating the difference, we want to divide by total population
  */
 export default function dataCalculator(data: DataPerGeoIdPerDate,
                                        range: [string, string],
```
shifucun, line 28: try to understand this function. Does this do the operation on a range of dates between range[0] and range[1]. Why not let the function do the stuff on a single date and let the client do the for loop?

covid19-dashboard/dc-covid19/src/DataCalculator.ts
edumorlom: date0 is updated on iteration.

`while (date0 <= date1){ date0++ }`

covid19-dashboard/dc-covid19/src/DataCalculator.ts
edumorlom: DataCalculator now only calculated the 'difference', 'increase', or absolute.
After DataCalculator finishes, then we calculate per-capita on the previous output.


**Should I separate absolutePerCapita and perCapita into in the config.json file?:**

```
'absolutePerCapita': ['absolute', 'perCapita'],
'perCapita: ['difference', 'perCapita']
```

covid19-dashboard/dc-covid19/src/DataCalculator.ts
edumorlom: This function no longer exists.

covid19-dashboard/dc-covid19/src/DataCalculator.ts
edumorlom: Done! The panel will now iterate through it range[0] to range[1]. This function no longer exists.

PR #13, opened 2020-06-24
per 10,000 people is now calculated as part of metadata instead of pe…

edumorlom:
Previously perCapita calculation would be in charge of multiplying by 10,000 people.
Now, metadata will multiply by 10,000 instead. Since it's really only used for on-hover.

We should keep per-capita to get per 1 person and any other calculation can be done elsewhere.


covid19-dashboard/dc-covid19/src/MetaDataGenerator.ts
```diff
@@ -1,123 +1,178 @@
...
-    if (contentType === 'perCapita' || contentType === 'absolutePerCapita') {
+const getOnHoverInfoText = (geoId: string,
+                            regionName: string,
+                            date: string,
+                            value: number,
+                            population: number,
+                            absolute: number,
+                            label: string,
+                            calculationType: string[]) => {
+    let disclaimer = ""
```
shifucun, line 44: Please add semicolon: https://google.github.io/styleguide/jsguide.html#formatting-semicolons-are-required

for all statements

covid19-dashboard/dc-covid19/src/MetaDataGenerator.ts
```diff
@@ -1,123 +1,178 @@
...
+    let onHoverInfo = [`${numberWithCommas(value)} ${label}`]
+
+    // If it's New York City or Kansas City. Show a disclaimer.
+    // These are NYT exceptions.
+    if (['geoId/3651000', 'geoId/2938000'].includes(geoId)) {
```
shifucun, line 50: make this a constant at the top of the file or, if shared, in a common module.

covid19-dashboard/dc-covid19/src/MetaDataGenerator.ts
```diff
@@ -1,123 +1,178 @@
...
+        disclaimer = `All counties in ${regionName}` +
+            'are combined and reported as one.'
+    }
+
```
shifucun, line 54: add a block comment:

// build the on hover information.

covid19-dashboard/dc-covid19/src/MetaDataGenerator.ts
```diff
@@ -1,123 +1,178 @@
...
-        onHoverInfo = [`${numberWithCommas(value)} ${label}`]
+            `Percent increase: ${percentIncrease}%`,
+            `Absolute increase: ${absoluteWithCommas} ${label}`
+        ]
     }
 
-    if (disclaimer) return [disclaimer, ...onHoverInfo]
-    else return onHoverInfo
+    if (disclaimer) {
+        return [disclaimer, ...onHoverInfo]
```
shifucun, line 74: should move this to line 51 and early return there

covid19-dashboard/dc-covid19/src/MetaDataGenerator.ts
```diff
@@ -1,123 +1,178 @@
...
-            const iterativeBelongingRegion: string = geoIdToName[geoId]?.[1]
-            const iterativeValue: number = dateToGeoIdToValue[date]?.[geoId] || 0
-            const iterativeAbsoluteIncreaseValue: number = absoluteIncrease[date]?.[geoId] || 0
-
-            // If the region not a State, append the State to the name. Example: "Miami, Florida"
-            if (iterativeBelongingRegion.includes("geoId"))
-                iterativeRegionName += ", " + geoIdToName[iterativeBelongingRegion]?.[0]
+            let regionName: string = geoIdToName[geoId]?.[0] || geoId
+
+            regionName.replace(" County", "")
```
shifucun, line 141: make "County", "Parish", "Borough" constant, and you will for sure expand the content in the future.

And use a for loop (or fancier way) to replace

covid19-dashboard/dc-covid19/src/MetaDataGenerator.ts
```diff
@@ -1,123 +1,178 @@
-import {numberWithCommas} from "./Utils";
+/**
```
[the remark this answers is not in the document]
shifucun: should be exact copy of this one: https://github.com/datacommonsorg/website/blob/master/static/js/gni.js#L1-L15

PR #15, opened 2020-06-24
Refactor

edumorlom:
Some minor refactoring to variable names.
Adding keys on .map iterations.
Dependecy update.
Nothing major.


covid19-dashboard/dc-covid19/src/Utils.ts
```diff
@@ -25,15 +25,15 @@ import moment from "moment";
  * @param deltaDays
  */
 const getRealISODatesFromArrayOfDeltaDays = (date: string, deltaDays: number[]): string[] => {
-    if (!date) return []
+    if (date === undefined || date === null) return []
```
shifucun, line 28: can just do !date ?

PR #23, opened 2020-07-21
Main file for the server-side in Flask

edumorlom:
This contains all the apis in the front-end. Queries data from DataCommons KG.

/*
/all-data (returns both state and county data)
/state-data (only returns state data), used for speed

The idea is that state-data and all-data are downloaded in parallel, but state-data should download much faster than all-data.

When state-data has been downloaded, the front-end will display the data. (The main page only has State data).
Once all-data finishes downloading, all-data overrides state-data and both are stored in local cache.


beets:
I'm about to start reviewing the code, but at a high level, why download state data twice? It should be one call for state, and one for county.

Best option would be, as discussed earlier, to limit the amount of data downloaded to just what's required for display. e.g.
/data/dcid -- data required for dcid - returning data for the dcid.

covid19-dashboard_NEW/main.py
```diff
@@ -0,0 +1,360 @@
...
+    # Adds HTTP headers for browser to store cache.
+    response = local_cache(data)
+    return response
+
+
+def get_data(counties=False):
+    """
+    Requests cases and deaths from the DC, and performs calculations.
+    :param counties: Whether we want both counties and states or just states.
+    :return: geoId->calculated data for both cases and deaths.
```
beets, line 122: would be great to document the format of the returned dict

covid19-dashboard_NEW/main.py
```diff
@@ -0,0 +1,360 @@
...
+        }
+
+    # Return a combination of cases + deaths + metadata.
+    return output
+
+
+def get_population():
+    """
+    Returns the total population of each region. geoId->population"
+    :return: geoId->population
```
beets, line 181: here as well

covid19-dashboard_NEW/main.py
```diff
@@ -0,0 +1,360 @@
...
+    # Kansas City.
+    output['geoId/2938000'] = 491_918
+
+    return output
+
+
+def get_places(return_counties=False):
+    """
+    Returns a list of places in the United States.
+    :param return_counties: whether to return information
```
beets, line 239: is the return value a list or dictionary of tuples?

covid19-dashboard_NEW/main.py
```diff
@@ -0,0 +1,360 @@
...
+    for counties+states or just states.
+    :return: geoId -> (name, containedIn, typeOfRegion)
+    """
+
+    # Get US State data.
+    response: dict = send_request(DC_SERVER + "node/places-in", {
+        "dcids": ["country/USA"],
+        "placeType": "State"
+    },
+                                  api_key=API_KEY)
```
beets, line 249: could you run pylint on this file?

covid19-dashboard_NEW/main.py
```diff
@@ -0,0 +1,360 @@
...
+                states[geo_id] = (name, 'country/USA', "State")
+
+    # Get US County data.
+    response: dict = send_request(DC_SERVER + "node/places-in", {
+        "dcids": state_dcids,
+        "placeType": "County"
+    },
+                                  api_key=API_KEY)
+
+    # Keep track of what county each state belongs to.
```
beets, line 275: counties belong to states, not the other way round :)

covid19-dashboard_NEW/main.py
```diff
@@ -0,0 +1,360 @@
...
+
+    # If param return_counties == True, get county data.
+    if return_counties:
+        for geo_id in response:
+            if 'out' in response[geo_id] and response[geo_id]['out']:
+                if 'value' in response[geo_id]['out'][0]:
+                    name: str = response[geo_id]['out'][0]['value']
+                    # Get rid of trailing strings in names such as County.
+                    for region_type in REGION_TYPES:
+                        name: str = name.replace(" " + region_type, "")
```
beets, line 301: you can just use rstrip() or strip() here. leaving the whitespace shouldn't effect display though.

btw, do we really include trailing spaces in our county names? that's in the dc graph now? please file a bug in the data repo so we fix it in the graph for everyone.

covid19-dashboard_NEW/main.py
```diff
@@ -0,0 +1,360 @@
...
+                    # The geoId where this region belongs to.
+                    belongs_to: str = county_to_state[geo_id]
+                    counties[geo_id] = (name, belongs_to, "County")
+
+        # NYT exceptions.
+        counties['geoId/3651000'] = ("New York City", "geoId/36", "County")
```
beets, line 307: are these missing in the dc graph? what's the exception that's being handled here.

covid19-dashboard_NEW/main.py
```diff
@@ -0,0 +1,360 @@
...
+
+    return {**states, **counties}
+
+
+def get_stats_by_date(place: list, stats_var: str) -> dict:
+    """
+    Queries and returns the data in the appropriate format.
+    :param place: list of dcids to get data for
+    :param stats_var: the Data Commons stats_var to query data for
+    :return: dictionary where every day contains several regions,
```
beets, line 318: please be more formal when describing the return values. what would be helpful for dictionaries is the keys in the dict, and the values contained therein.

covid19-dashboard_NEW/main.py
```diff
@@ -0,0 +1,360 @@
...
+            continue
+        if "data" not in response[dcid]:
+            continue
+        data: dict = response[dcid]["data"]
+        for date in data:
+            output[dcid][date] = data[date]
+    return output
+
+
+def local_cache(data, minutes=100):
```
beets, line 341: i think a better name for this is add_browser_cache_headers_to_response or something to that effect. local_cache is too vague.

covid19-dashboard_NEW/main.py
```diff
@@ -0,0 +1,360 @@
...
+from calculate_data import calculate_data
+
+config = dict(DEBUG=True,
+              CACHE_TYPE="filesystem",
+              CACHE_DIR='/tmp')
+
+app = Flask(__name__, static_folder='./build')
+Compress(app)
+
+CORS(app)
```
beets, line 34: Is CORS required? I think all requests are coming to this server, so let's avoid adding CORS if we can.

covid19-dashboard_NEW/main.py
edumorlom: The KG has several alternative names including different languages, so it does contain "Miami-Dade" alone, but I need to figure out how to request it using APIs. Will look into it.

covid19-dashboard_NEW/main.py
edumorlom: NYT combines several counties into one larger county.
These aren't real counties, but we're treating them as such.
https://github.com/nytimes/covid-19-data#geographic-exceptions

covid19-dashboard_NEW/main.py
edumorlom: i added detailed types such as Dict[str, Dict[str, str]] and documented the code better.
Also added README.md with more information.

covid19-dashboard_NEW/main.py
edumorlom: Whoops thanks!

covid19-dashboard_NEW/main.py
edumorlom: Removed CORS.

covid19-dashboard_NEW/main.py
edumorlom: Done!

covid19-dashboard_NEW/main.py
edumorlom: done

covid19-dashboard_NEW/main.py
edumorlom: i added detailed types such as Dict[str, int] and documented the code better.

edumorlom:
> I just noticed the folder name is _NEW. Do you want to just replace the existing files instead of making a new folder? There are also other things to clean up, e.g. package-lock.json, etc.

I moved the files to the other directory, there are two commits.
1. For code changes.
2. For moving the files to the other directory.

edumorlom:
Now there are two apis.
**/api/state-data
/api/county-data**

previously:
**/api/state-data
/api/all-data (which included state-data)**

covid19-dashboard/main.py
```diff
@@ -12,201 +12,391 @@
...
+# For example, "Miami-Dade County" becomes "Miami-Dade".
+PLACE_TYPES = ["County", "Parish", "Borough"]
+
+# COMMONLY-USED TYPES.
+OutputDataType = Dict[str, Dict[str, int] or Dict[str, str]]
+GeoIdToDataType = Dict[str, Dict[str, int]]
+RegionToInfoType = Dict[str, Tuple[str, str, str]]
+
+
+@app.route('/<path:path>', methods=['GET'])
```
beets, line 62: This can be done using app engine, bypassing flask:
https://github.com/datacommonsorg/website/blob/master/server/app_prod.yaml

Not a good practice to be able to send all files from root, since there might be sensitive files that get accidentally shipped.

Btw, is this deployed? This doesn't seem to work.

covid19-dashboard/main.py
```diff
@@ -12,201 +12,391 @@
...
+            # Perform XYZ calculations.
+            calculated_cases = calculate_data("Cases", cases, region_pop)
+        if geo_id in all_deaths:
+            deaths = all_deaths[geo_id]
+            # Perform XYZ calculations.
+            calculated_deaths = calculate_data("Deaths", deaths, region_pop)
+
+        name: str = ""
+        contained_in_geo_id: str = ""
+        place_type: str = ""
```
beets, line 170: It's generally not good practice to override a function parameter like this.

covid19-dashboard/main.py
```diff
@@ -12,201 +12,391 @@
...
+# DataCommons Statvars to be used.
+INCREMENTAL_CASES_STATVAR = (
+    "IncrementalCount_MedicalConditionIncident_COVID_19_ConfirmedOrProbableCase"
+)
+INCREMENTAL_DEATHS_STATVAR = (
+    "IncrementalCount_MedicalConditionIncident_COVID_19_PatientDeceased")
+
+# The type of sub-regions. Used to remove trailing from name.
+# For example, "Miami-Dade County" becomes "Miami-Dade".
+PLACE_TYPES = ["County", "Parish", "Borough"]
```
beets, line 54: Hm, I'm not sure if this is necessary to remove. Was this a requested feature?

edumorlom: I added it because the names were getting too long.

Northumberland County, Pennsylvania

But i will remove and will look for a way to request an alternate name from DC KG.

covid19-dashboard/main.py
edumorlom: No need to overwrite it  since this function only gets data based on place_type input parameter, so all should be of the same place_type. Not necessary to retrieve it again.

edumorlom: Removed overwrite. Thanks

covid19-dashboard/main.py
edumorlom: app_static_folder = ./build/
The build folder only contains static files from output 'react run build', these include CSS, JS, images and HTML.

Unless there is some way of accessing an outside directory.
Let's discuss this tomorrow, I'm not sure how else to send the static files then.

edumorlom: PS. It is deployed.
What URL are u trying to access?

edumorlom:
Changes:
1. /api/countries is now available.
2. I made the server stat_var independent. Any stat_var should work.
3. Changes you requested.
4. More documentation.
5. New _get_place_names function, previously duplicates in code.

edumorlom:
placeType and containedIn are the properties that DC uses.
I renamed variables to be consistent.

covid19-dashboard/README.md
```diff
@@ -1,15 +1,61 @@
...
-2. Set Up an App Engine Instance.
-3. Login to your App Engine Instance using the Google Cloud Command Line Tools.
-4. Execute ``gcloud app deploy``.
-5. Execute ``gcloud app deploy cron.yaml``.
 
+#### Server APIs:
+The APIs are based on a ```place_type```.
+
+**api/country-data**: returns the data for all Countries in the world with WHO COVID-19 data.
+ - downloads in a few ms.
```
beets, line 14: These statements are hard to stand on it's own - with what connection speed, what was the server processing time? etc.

What stands is the size of the returned payload, and perhaps server wall time. Just payload size should suffice though.

covid19-dashboard/main.py
```diff
@@ -12,201 +12,488 @@
...
             continue
-        data: dict = response[dcid]["data"]
+
+        county_geo_id: str = value["place"]
+        belongs_to_geo_id: str = value["dcid"]
+        county_to_state[county_geo_id]: str = belongs_to_geo_id
+
+    # If the user didn't request County data, they requested State data.
+    # Return state data.
+    if place_type != "County":
```
beets, line 421: this statement can be moved up to save some computation.

covid19-dashboard/README.md
edumorlom: Fixed.

covid19-dashboard/main.py
edumorlom: Fixed, thanks!

beets:
Looks good - please make sure to make sure your branch is up to date so we can merge.

edumorlom:
Ok it's up to date!

PR #25, opened 2020-07-21
send_request function: used to get data from DC KG REST apis

edumorlom:
This function simply requests data from a given API.
In my case I use it to retrieve data from the DC Rest APIs .


edumorlom:
I was initially using the Python APIs but it was missing the stat_var function so I was suggested to migrate to REST API since it's the most-up-to-date, but after talking to Bo recommended I give the Python API another try now that stat_var is available.

Will make a pull request soon to migrate to Python API.

beets:
Ok, we can do that as a followup.

edumorlom:
@beets I don't think interns are authorized to merge the PR.

beets:
merged!

PR #31, opened 2020-07-31
Gcloud updates + dependencies

edumorlom:
Currently we are manual scaling to 1 instance.
We haven't implemented memcache yet, and so 1 instance helps since there isn't a lot of traffic.


beets:
Why is memcache required to keep autoscaling? Even if computations are repeated, it's only once per instance, which is better than getting bottlenecked on one instance... I'd keep autoscaling if possible.

beets:
Any updates here?

edumorlom:
I've reverted to autoscaling but upgraded server to python3.8

PR #35, opened 2020-08-07
Initial PR for dashboard 2.0

edumorlom:
Client Diagram will help visualize what is going on!
https://docs.google.com/drawings/d/18YGj-rx9O8Bz6n-0o7mFy00sejmigJRrPrsQi0R9rN0/edit?usp=sharing

There are 3 main components.
The rest of the components are meant to help modulize the code.

Most of the logic lives here!!:
**App.tsx
DataTable.tsx**

The table headers, dataKeys, etc are all taken from Content.json.
App configuration can be found under Configuration.json

I've tried my best to document everything. Let me know if there are too many comments.
**I apologize for this big PR, other PRs will be smaller from now on. Wasn't sure how to break this PR into features since it's all one big feature**

PR #36, opened 2020-08-10
makes all stats cumulative and then we calculate incremental

edumorlom:
Because most stats on Data Commons are cumulative, in an effort to make the dashboard compatible with other stats, I now calculate the incremental values myself. So i switched to a different stat_vars on config.py.

Also included _difference() function in charge of calculating incremental values. I will add unit tests soon.

ALSO: in config.py, it adds recovered stat_var


beets:
I'll review this after we get PR #35 merged. There are still outstanding comments there.

edumorlom:
Now instead of having individual /data/Country, /data/County, etc... I have @app.route("/api/data/<string:geo_id>")

covid19-dashboard/calculate_data.py
```diff
@@ -18,6 +18,31 @@
 DateToValueListType = List[Tuple[str, float]]
 
 
+def _difference(date_to_value: DateToValueListType,
```
beets, line 21: please add documentation for the function

covid19-dashboard/main.py
```diff
@@ -498,5 +487,13 @@ def add_browser_cache_headers_to_response(data: Dict,
     return response
 
 
+def _place_info_factory(name, contained_in, place_type):
```
beets, line 490: this is technically not a factory... why not just call it _create_place_info.
also probably worth making it a type.

covid19-dashboard/calculate_data.py
edumorlom: fixed.

covid19-dashboard/main.py
edumorlom: i made it into a class and then do __dict__. is  that ok?

edumorlom:
ready for re-review! thanks

PR #37, opened 2020-08-17
adds required html, css files

edumorlom:
Adds required public files such as index.html.
Adds CSS files.
Removes helper_functions.py, no longer needed.

PR #38, opened 2020-08-17
Fixes a bug onClick and adds caching function

edumorlom:
1. Adds function to cache data.
2. Fixes a bug with onClick in data table where all places were clickable even if they didn't have subregions.


covid19-dashboard/src/DataTable.tsx
```diff
@@ -212,15 +212,20 @@ export default class DataTable extends React.Component<
...
-        <tr key={index}
-            className={clickableClass}
-            onClick={() => goToPlace(place.geoId, place.getSubregionType())}>
-          <th>{tableRanking}</th>
-          <th>{placeFullName}</th>
-          {thValues}
-        </tr>
+          <tr key={index}
+              className={clickableClass}
+              onClick={onClick}>
```
beets, line 224: instead, only register the onclick if there is a subregion. and... add a class if it's clickable so you can give a click affordance to users, e.g changing the cursor and/or underline if it's clickable

edumorlom:
when does something get read from the cache? **The caching is used so that when the user navigates to a County, the data is already available and the page loads instantly. Otherwise, the County data takes a few secs to download on page load.**

why isn't placetypetoshow also stored in the cache? **fetchData() is downloading the placeTypeToShow data and also storing it in the cache (but also processing the data) so I don't think there is a need to re-download the data in the cacheData(), which is why I remove placeTypeToShow from the cacheData() array.
So in other words, fetchData() processes the data and caches it, cacheData() only caches the data.**

i understand wanting to preemptively load the data, but it's not being used so you're just relying on the browser to cache the response. the previous iteration actually processed the response, which is not happening anymore. **The data processing was unnecessary since we aren't showing those places. For example, if I'm in the US why would I want to process data for France**

is there an issue with speed in the first place? is this a premature optimization? **So I noticed that processing all the data caused some lag in rendering, specially with animation because the app was re-rendered for each placeType. If I'm viewing Countries, I don't really need States or Counties**.

covid19-dashboard/src/DataTable.tsx
edumorlom: Please take a look at the last commit, let me know if that's what you meant by "only register the onClick" and "add a class if it's clickable". Thanks!

Sorry, I had to force-push because I made a mistake and reverted.

covid19-dashboard/src/DataTable.tsx
```diff
@@ -195,33 +184,24 @@ export default class DataTable extends React.Component<
...
-        onClick = () => goToPlace(place.geoId, subregionType)
+        placeFullName += `, ${place.parentPlace?.name}`;
       }
 
       return (
         <tbody key={index}>
-          <tr key={index}
-              className={clickableClass}
-              onClick={onClick}>
+          <tr className={subregionType ? 'clickable' : ''}
```
edumorlom, line 202: It this what you meant by "add a class?"

covid19-dashboard/src/DataTable.tsx
```diff
@@ -195,33 +184,24 @@ export default class DataTable extends React.Component<
...
+              {...(subregionType && {
```
edumorlom, line 203: Is this what you meant by "only register onClick if"

covid19-dashboard/src/DataTable.tsx
edumorlom: clickable is the CSS class

covid19-dashboard/src/DataTable.tsx
edumorlom: if there is a subregion, then onClick is registered, otherwise no onClick is passed.

beets: yes, this is what i meant. any css changes to add here? updated color or font decoration, cursor: pointer, etc to denote that something is clickable vs other non-clickable things?

covid19-dashboard/src/App.tsx
```diff
@@ -71,49 +71,84 @@ class App extends React.Component<AppPropsType, AppStateType> {
...
+    // We don't really care about this data, we only wanted to cache it.
+    // None the less, we still store it in the background because
+    // it can help us search when using the search bar.
+    Promise.all(cachePromises).then(responses => {
+      responses.forEach(response => {
+        response.json().then(json => {
+          Object.entries(json).forEach(([geoId, value]) => {
+            if (geoId in this.places) {
+              // Store the timeSeries data for the geoId.
+              this.places[geoId].keyToTimeSeries = value as KeyToTimeSeriesType
```
beets, line 145: yes, now this is a cache.

covid19-dashboard/src/DataTable.tsx
edumorlom: It was added along the HTML file. There is clickable:hover in the index.css file.

https://github.com/datacommonsorg/tools/pull/37

edumorlom:
If you think this PR is good to go, I think it's ready to merge.

beets:
See my comment above about css changes to denote clickable state

edumorlom:
Hey @beets , the CSS class already existed prior to this PR.

The changes to the CSS can be found in this PR -- which includes all HTML and CSS files.

https://github.com/datacommonsorg/tools/pull/37/

beets (approved):
great, thanks for pointing out the old PR!

PR #40, opened 2020-08-31
Covid19 app engine service

edumorlom:
Creates covid19.yaml file with covid19 service. The app engine service is in charge of sending static files instead of Flask.

When /covid19/ is requested, redirect user to covid19 service.


edumorlom:
I'm not sure if this is the way things should be done.
Some initial feedback would be good.

beets:
beyond the structure of these files, i believe you had run into other issues setting this up and wanted help. what were the issue, what have you tried, and have you tested this setup? how does it work, and what did you find? it would be good to know what you want feedback on.

edumorlom:
I spoke to Bo, and the way we decided to solve the /covid19 issue is to have a simple HTTP redirect on /covid19 that redirects to /covid19/

This PR moves all files to covid19 directory, and has a redirect on default/main.py that will redirect to /covid19/

Now /covid19 and /covid19/ should both work.

edumorlom:
Will close until other PR is merged.

PR #41, opened 2020-09-01
Go module for mean imputation

eftekhari-mhs:


imputation/src/mean/mean_test.go
```diff
@@ -0,0 +1,163 @@
...
+    series3["1998"] = 28
+    
+    want = make(map [string]float64)
+    want["2003"] = 42.5
+    want["2002"] = 18.9
+    want["1998"] = 28
+   
+    
+    var mean float64 = 0
+    for _,v := range series3{
```
jeffreyoldham, line 51:
```suggestion
    for _, v := range series3 {
```

imputation/src/mean/mean_test.go
```diff
@@ -0,0 +1,163 @@
...
+    want["2003-02"] = 42.5
+    want["2002-02"] = 18.9
+    want["2003-04"] = 28
+   
+    
+    mean = 0
+    for _,v := range series4{
+        mean += v
+    }
+    mean = mean/float64(len(series4))
```
jeffreyoldham, line 87: Let's define a function to compute the mean for use within this file.  This follows the heuristic of writign a function if code is need twice.

imputation/src/mean/mean_test.go
```diff
@@ -0,0 +1,163 @@
+package main
+
+import (
+    "testing"
+)
+
+func TestMean(t *testing.T) {
+    // test 1 empty series
```
jeffreyoldham, line 8: Unlike all the programming languages I know, Go prefers table-driven tests.  Q.v., https://dave.cheney.net/2019/05/07/prefer-table-driven-tests.  Let's add a todo to update this code to be more table driven.

```
TODO(eftekhari-mhs): Refactor to table-driven tests.
```

imputation/src/mean/mean_test.go
```diff
@@ -0,0 +1,163 @@
...
+        if vw, ok := want[k]; !ok || vw != v {
+            t.Errorf("Value mismatch: fillMean() = %v, want %v", got, want)
+        }
+    } 
+    
+    
+    //test 4 month gap
+    series4 := make(map [string]float64)
+    series4["2003-02"] = 42.5
+    series4["2002-02"] = 18.9 //edge case : 2002-01 fails currently
```
jeffreyoldham, line 75: Let's make the comment into a todo, making it easier to find later.

imputation/src/mean/mean_test.go
```diff
@@ -0,0 +1,163 @@
...
+        mean += v
+    }
+    mean = mean/float64(len(series4))
+    
+    want["2002-04"] = mean
+    want["2002-06"] = mean
+    want["2002-08"] = mean
+    want["2002-10"] = mean
+    want["2002-12"] = mean
+    want["2002-08"] = mean
```
jeffreyoldham, line 94: Why is this repeated?

imputation/src/mean/mean_test.go
```diff
@@ -0,0 +1,163 @@
...
+    
+    for k,v := range got{
+        if vw, ok := want[k]; !ok || vw != v {
+            t.Errorf("Value mismatch: fillMean() = %v, want %v", got, want)
+        }
+    } 
+    
+    
+    //test 4 month gap
+    series4 := make(map [string]float64)
```
jeffreyoldham, line 73: The pattern seems to be:
Create an input time series.
Create an output time series by adding some values to the input time series. 

I think we can simplify the code like this:
```
series4: TimeSeries{
  "2003-02": 42.5,
  "2002-02": 18.9,
  "2003-04": 28,
}
want = createOutput(series4, ["2002-04", "2002-06", "2002-08", "2002-10", "2002-12"])
```

```
func createOutput(ts TimeSeries, desiredValues) TimeSeries {
  out := make(TimeSeries, len(ts) + len(desiredValues))
  for k, v := range ts {
    out[k] = v
  }
  m := mean(ts)
  for _, k := range(desiredValues) {
     out[k] = m
  }
  return(out)
}
```

I think this works for all five test instances.

imputation/src/mean/go.mod
```diff
@@ -0,0 +1,3 @@
+module mean
```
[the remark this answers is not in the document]
eftekhari-mhs: good point, I think it's better to support more, I will change the module name

imputation/src/mean/mean.go
```diff
@@ -0,0 +1,76 @@
...
+
+	duration := 1000 //a large number
+
+	sort.Strings(keys)
+	for i := 0; i < len(keys)-1; i++ {
+		start, _ := time.Parse(dateFormat, keys[i])
+		end, _ := time.Parse(dateFormat, keys[i+1])
+
+		// calculate total number of days
+		if delta := int(end.Sub(start).Hours() / 24); duration > delta { //TODO: instead of min use GCD
```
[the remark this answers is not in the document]
eftekhari-mhs: We should look at all consecutive points. We can stop when duration is 24 hrs, because we're assuming this is the minimum gap possible.

imputation/src/mean/mean.go
```diff
@@ -0,0 +1,76 @@
...
+	case 7:
+		parseFormat = monthfmt
+	}
+
+	yStep, mStep, dStep := diff(keys, parseFormat)
+
+	fmt.Printf("Step is equal to : %v years, %v months, %v days \n", yStep, mStep, dStep)
+
+	startDate, _ := time.Parse(parseFormat, keys[0])
+	endDate, _ := time.Parse(parseFormat, keys[len(keys)-1])
```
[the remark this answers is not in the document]
eftekhari-mhs: Ah, right, I can also find the min and max in linear time which might help performance.

imputation/src/mean/mean_test.go
eftekhari-mhs: Thank you so much!!! This will help readability and future testing a lot.

imputation/tsImpute.go
```diff
@@ -0,0 +1,102 @@
...
+	switch len(keys[0]) {
+	case 4:
+		parseFormat = yearfmt
+	case 7:
+		parseFormat = monthfmt
+    case 10:
+        parseFormat = dayfmt
+	}
+    
+    if parseFormat == "err" {
```
jeffreyoldham, line 82:
```suggestion
	var parseFormat string
	switch len(keys[0]) {
	case 4:
		parseFormat = yearfmt
	case 7:
		parseFormat = monthfmt
    case 10:
        parseFormat = dayfmt
	}
    
    if parseFormat == "" {
```

I think this suggested code is obscure, but it is the preferred Go idiom.  `var parseFormat` creates a variable without assigning a value so it gets the zero value, which is "" for strings.

Using an empty string is better than using "err" because it is clear that the empty string cannot be a key for a time series.  <-- That explanation is not very good or explanatory.

imputation/tsImpute.go
```diff
@@ -0,0 +1,102 @@
...
+			month = int(M2 - M1)
+			day = int(d2 - d1)
+		}
+	}
+
+	return year, month, day
+}
+
+// FillMean returns TimeSeries with missing datapoints with value = mean(value of existing points).
+// TODO(eftekhari-mhs): Add error to the return values and handle error cases.
```
jeffreyoldham, line 59:
```suggestion
// TODO(eftekhari-mhs): Handle error cases.
```

imputation/tsImpute_test.go
```diff
@@ -0,0 +1,173 @@
...
+//TODO(eftekhari-mhs): Refactor to table-driven tests: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests
+
+import (
+    "testing"
+    "errors"
+)
+
+func computeMean(s TimeSeries) (float64, error){
+    if len(s) ==0 {
+        //TODO err handling: divison by zero
```
jeffreyoldham, line 12:
```suggestion
```

imputation/tsImpute_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+//TODO(eftekhari-mhs): Refactor to table-driven tests: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests
+
+import (
+    "testing"
+    "errors"
+)
+
+func computeMean(s TimeSeries) (float64, error){
+    if len(s) ==0 {
```
jeffreyoldham, line 11: Please consider running "go fmt" on this file, too.

imputation/tsImpute_test.go
```diff
@@ -0,0 +1,173 @@
+package imputation
+
+//TODO(eftekhari-mhs): Refactor to table-driven tests: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests
+
+import (
+    "testing"
+    "errors"
+)
+
+func computeMean(s TimeSeries) (float64, error){
```
jeffreyoldham, line 10: In this particular function designed only for testing, it might suffice to not check for division by zero and just dying if given an empty time series.  This is because the test writer should know to not call the function on such a time series.  If e does, the test will crash or fail.  In regular user code where one has to handle all cases gracefully.

imputation/tsImpute_test.go
```diff
@@ -0,0 +1,173 @@
...
+func createOutput(ts TimeSeries, desiredValues []string) TimeSeries {
+  out := make(TimeSeries, len(ts) + len(desiredValues))
+  for k, v := range ts {
+    out[k] = v
+  }
+  m, err := computeMean(ts)
+  if err != nil{
+    return ts    
+  }
+
```
jeffreyoldham, line 32: If you take the above suggestion, this code becomes `m := computeMean(ts)`.

jeffreyoldham (approved):
Thanks.  After you address the comments, let's submit.  Thanks.

imputation/tsImpute.go
```diff
@@ -0,0 +1,102 @@
...
+
+The writter library first find the minimum gap of existing points and then add
+new data points with values according to the selected method.
+*/
+package imputation
+
+import (
+    "fmt"
+	"sort"
+	"time"
```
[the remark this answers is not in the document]
eftekhari-mhs: This is weird, I run  gotfmt -w on both files and copy paste the result to GitHub. Do you know why the format is still not correct? On my local machine it doesn't show odd spacing...

imputation/tsImpute_test.go
eftekhari-mhs: Is this because developer does the testing and not regular user?

imputation/tsImpute.go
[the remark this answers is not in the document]
jeffreyoldham: It probably has to do with tabs.  http://screen/3L7vf8H79AiS5dH.  Tabs can have different widths.  Also, they work best with fixed-width characters.

If you are using gofmt, let's assume this is the best we can do.

imputation/ts_impute_test.go
```diff
@@ -0,0 +1,112 @@
...
+func createOutput(ts TimeSeries, desiredValues []string) TimeSeries {
+	out := make(TimeSeries, len(ts)+len(desiredValues))
+	for k, v := range ts {
+		out[k] = v
+	}
+	m := computeMean(ts)
+
+	for _, k := range desiredValues {
+		out[k] = m
+	}
```
jeffreyoldham, line 25:
```suggestion
        if len(desiredValues) > 0 {
	  m := computeMean(ts)
	  for _, k := range desiredValues {
		out[k] = m
	  }
	}
```

This avoids division by zero in the mean function.

imputation/ts_impute_test.go
```diff
@@ -0,0 +1,112 @@
+package imputation
+
+import (
+	"testing"
+)
+
+func computeMean(s TimeSeries) float64 {
```
jeffreyoldham, line 7: // computeMeans computes the mean of the given non-empty time series's data.

jeffreyoldham (dismissed):
Great.  Let's get it in.

imputation/ts_impute_test.go
eftekhari-mhs: Do you mean if len(ts) > 0?

jeffreyoldham (approved):
Thanks.

imputation/README.md
```diff
@@ -0,0 +1 @@
+Go module for adding missing dates of a time series with value = mean
```
tjann, line 1: Please flush this out later

imputation/ts_impute.go
```diff
@@ -0,0 +1,103 @@
...
+	yearfmt  = "2006"
+)
+
+// The diff function returns the minimum gap (will change to GCD) from a given set of data points
+// that all have the same format.
+// TODO(eftekhari-mhs): add OK, ERR return values and handle errors.
+func diff(keys []string, dateFormat string) (int, int, int) {
+	year, month, day := 0, 0, 0
+
+	duration := 1000 //A large number.
```
tjann, line 36: Can doing using something like this https://yourbasic.org/golang/max-min-int-uint/

imputation/ts_impute.go
```diff
@@ -0,0 +1,103 @@
...
+
+	sort.Strings(keys)
+	for i := 0; i < len(keys)-1; i++ {
+		start, _ := time.Parse(dateFormat, keys[i])
+		end, _ := time.Parse(dateFormat, keys[i+1])
+
+		// Calculate total number of days.
```
tjann, line 43: Is this comment updated?

tjann (approved):
This is a hand waving approval.

PR #47, opened 2020-09-11
Small refactoring

edumorlom:
Some small code refactoring + adds package.json.

### DIFF
--- a/covid19-dashboard/src/Utils.ts
+++ b/covid19-dashboard/src/Utils.ts
@@ -54,14 +54,14 @@ export const Colors = (color: string): string => {
  * would display all counties in the USA.
  */
 export const goToPlace = (geoId?: string, placeType?: string): void => {
-  let newUrl = '/';
+  let newUrl = '/covid19/';
 
   // Redirect to a specific geoId.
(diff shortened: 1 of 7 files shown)


covid19-dashboard/src/Utils.ts
```diff
@@ -26,14 +26,20 @@ export const numberWithCommas = (num: number): string => {
 };
 
 /**
- * Gets the latest date in the for a a list of Iso-dates.
+ * Gets the min or max date in a list of ISO dates.
```
edumorlom, line 29: Revert back to getLatestDate

edumorlom: Can we do this in a separate PR so that can rename the function in the necessary files?

beets: perhaps this PR isn't complete then. I don't see the renamed function used in these files (or it's not obvious anyway). Could you just revert this change?

edumorlom:
I reverted the getMinMaxDate to getLatestDate.
Other changes remain

beets (approved):

PR #62, opened 2020-11-16
mcf browser - front end files

clincoln8:
The front end has the overall layout-
starting point: index.js 
App.js - drives the app and maintains the state of the loaded files and which node to load
----Header.js - contains the 'return home' button, 'choose file' button, and search bar
----DisplayNode.js - requests necessary info for the current node and sets up the page to display the node
--------TriplesTable.js
----Home.js - displays the current files, 'clear files' button, any parsing errors, and the list of subject nodes

src/utils.js has functions that are used by many of the src/ files, mostly relating to modifying the browser's url hash.
src/back-end/server-api.js has files to relay information from the backend files to the front end


shifucun (commented):
Some high level questions.

Is the App created from create react app? Can you add a README on how to do local testing and deploy this? I guess here need a package.json to install node modules?

mcf-browser/src/back-end/server-api.js
```diff
@@ -0,0 +1,133 @@
...
+  * is used as the jsx element id for the text containing the node id.
+  * Blue = dcid exists in the DC KG
+  * Purple = no dcid, local id is resolved in the local data
+  * Orange = no dcid, local id is unresolved
+  * Red = default, including dcid does not exist in KG
+  *
+  * @param {Node} target The node object whose element color needs to be found.
+  * @return {String} The appropriate display color for the node.
+  */
+async function getElemId(target) {
```
beets, line 102: This shouldn't return id's, but should return classes, and the classes should have semantic meaning (e.g. exists-in-kg, etc)

We can always set these colors in CSS, and it would be much easier to update in the future

mcf-browser/src/utils.js
```diff
@@ -0,0 +1,90 @@
...
+function goToHome() {
+  window.location.hash = fileHash;
+}
+
+/**
+ * Saves the file params as part of the current url hash so that the files
+ * remain in the url when navigating between nodes.
+ * @param {Array<String>} fileUrlList List of file urls to be saved in the hash.
+ */
+function setFileHash(fileUrlList) {
```
beets, line 58: should there be a corollary "getFileHash" here?

mcf-browser/src/utils.js
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import React from 'react';
+
+let fileHash = '#';
```
beets, line 18: why is this global?

mcf-browser/src/utils.js
```diff
@@ -0,0 +1,90 @@
...
+ * @param {String} fileUrl Url of the fileee to open.
+ */
+function openFile(fileUrl) {
+  if (fileUrl.startsWith('https')) {
+    window.open(fileUrl);
+  }
+}
+
+// spinning animation to demonstrate loading, used in DisplayNode and Home
+const LoadingSpinner = (props) => {
```
beets, line 76: please pull this out to a new file

beets (commented):
Mostly looks good! Some small comments, but also in general, please rename the js files to jsx if they contain react components or return JSX elements).

It would be great to run eslint on this too.

beets:
Please also add a README, and especially include instructions to run and deploy the app - thanks!

mcf-browser/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+After installing node_modules, you can run the tests for the back-end by running
+```sh
+$ npm run test
+```
+
+You can deploy a local version of the application to test the front-end with
+```sh
+$ npm start
+```
```
beets, line 83: Please add a note about deploying a new version as well (even if it's with github-pages -- any notes for us to deploy this in the future?)

beets: Note that we are now deploying the entire repo to github pages...
https://datacommonsorg.github.io/tools/imputation/

Could you add a note about this, and that we should try to move this elsewhere?

beets (commented):
Thanks for responding to the comments. Looks ok, but I see that since we have enabled github pages for this repo, our entire repo is being published. Could you resolve this last comment? At this point, we can document, but we might want to deploy this with the appengine instance in the future.

beets (approved):

beets:
Thanks for the updates!

PR #72, opened 2020-12-17
deploy updates

beets:
add set -e to bash scripts to fail on error
update path back to build
add build to .gitignore


edumorlom (approved):
Looks good thanks! i didnt know about "set -e"

PR #82, opened 2021-01-11
Modify CSV-to-BT pipeline to write completion file

pradh:
This PR is the first step to simplify the BT automation flow.

Right now the dataflow pipeline has no way to inform back on completion. Thus, the current flow uses a more complex Airflow system to achieve that. The flow in the figure below involves 2 cloud functions, 1 airflow dag, a dataflow pipeline and multiple sync points:

Screen Shot 2021-01-11 at 1 35 05 PM

Sushant mentioned that the reason was that dataflow could not signal completion. This change modifies the Dataflow Java code to use the new API [BigtableIO.WriteWithResults](https://beam.apache.org/releases/javadoc/2.19.0/org/apache/beam/sdk/io/gcp/bigtable/BigtableIO.WriteWithResults.html) which provides the ability to wait for writes to complete and then write a single completion file. As a result, we can have a simpler flow like below:

image

Modifications to the go cloud function to follow next.

I tested this with a manual run (documented in the README.md). Adding a full integration test, using Cloud Bigtable emulator, is still a TODO.


shifucun (commented):
do you need to check in the java class files?

pradh:
> do you need to check in the java class files?

Don't think so. But I messed up my local branch, so I'll push this and do another PR to drop those.

Thanks for the review!

PR #87, opened 2021-01-16
Clean up legacy stuff and Improve Csv2BT Dataflow Validation

pradh:


shifucun (approved):
thanks for the change!

PR #93, opened 2021-03-24
Reduce nodes in datcom-store project to 298.

pradh:
The max BT node limit for datcom-store is 300. Of which 2 are reserved for branch cluster, so reduce the main cluster's limit by 2.


shifucun (approved):

pradh:
Thanks.

PR #94, opened 2021-03-24
Fix lint errors in covid19-dashboard

beets:
Tested with:
npm run lint
npm install
npm run build

And ran a local server: ./run_locally.sh


shifucun (approved):
thanks for the cleanup!

PR #110, opened 2022-02-04
Improve BT automation

pradh:
List of changes:

1. Set usePublicIps=false.  This required enabling Private Network in us-central1 subnet (screenshot below).  I validated it by running the `test.sh` tool, but was somewhat puzzled to not find the usePublicIps option listed on the [job page](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2022-02-03_15_52_03-15865834863613240020?project=google.com:datcom-store-dev).  However, I verified that the running VMs were using private IP (screenshot below).

2. Pass --region='us-central1' explicitly (it seems to be an upcoming required arg), and also given the private network requirement, it seems better to set.

3. Add a tool to deploy the Cloud Function to prod.

image

image

### DIFF
--- a/bigtable_automation/gcf/go.mod
+++ b/bigtable_automation/gcf/go.mod
@@ -1,6 +1,6 @@
-module btcachegeneration
+module example.com/btcachegeneration
 
-go 1.15
+go 1.13
 
 require (
 	cloud.google.com/go/bigtable v1.6.0
--- a/bigtable_automation/gcf/README.md
+++ b/bigtable_automation/gcf/README.md
@@ -82,8 +82,12 @@ go build main.go
 
 ## Deployment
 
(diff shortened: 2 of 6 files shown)


bigtable_automation/gcf/go.mod
```diff
@@ -1,6 +1,6 @@
-module btcachegeneration
+module example.com/btcachegeneration
 
-go 1.15
+go 1.13
```
shifucun, line 3: why change this to older version?

shifucun (approved):

bigtable_automation/gcf/go.mod
pradh: The runtime was using 1.13, so I just made this one consistent with that.

PR #131, opened 2022-07-22
Update MCF Browser to use new DC API

cwcrystal8:
Repo has the following changes:

1. Change the following three API calls to use v1 DC REST API
  - Check if a DCID exists in the DC KG
  - Get all property labels for a DCID
  - Get a property value for a given DCID and property label

2. Rebuild static files


mcf-browser/src/back-end/utils.js
```diff
@@ -61,13 +70,12 @@ async function getRemotePropertyLabels(dcid) {
 async function getRemotePropertyValues(dcid, label, isInverse) {
   const direction = isInverse ? 'in' : 'out';
   const targetUrl =
-      (API_ROOT + '/node/property-values?limit=500&dcids=' + dcid +
-       '&property=' + label + '&direction=' + direction);
+      (API_ROOT + '/v1/property/' + direction + '/' + label +
```
juliawu, line 73: We're updating the v1 REST API signature to /v1/property/values/{direction}/{dcid}/{property}
@shifucun to confirm: We should update to the new signature now, right?

mcf-browser/src/back-end/utils.js
```diff
@@ -103,10 +111,13 @@ function getValueFromValueObj(valueObj) {
  *     Data Commons Knowledge Graph.
  */
 async function doesExistsInKG(dcid) {
-  const url = API_ROOT + '/node/triples?dcids=' + dcid + '&limit=1';
+  const url = API_ROOT + '/v1/property/out/typeOf/values/' + dcid;
```
juliawu, line 114: Same comment here: update to /v1/property/values/out/ + dcid + /typeOf

mcf-browser/src/back-end/utils.js
cwcrystal8: Do you know when the new signature update is happening? Last time I spoke to @shifucun, I think he said to keep using the old signature for now.

mcf-browser/src/back-end/utils.js
```diff
@@ -42,10 +42,19 @@ const ERROR_MESSAGES =
...
+  const outPropertyLabels = await fetch(outTargetUrl)
+      .then((res) => res.json())
+      .then((data) => data.properties);
+
+  // Get inward property labels
+  const inTargetUrl = API_ROOT + '/v1/properties/in/' + dcid;
+  const inPropertyLabels = await fetch(inTargetUrl)
       .then((res) => res.json())
-      .then((data) => JSON.parse(data.payload)[dcid]);
+      .then((data) => data.properties);
```
shifucun, line 55: You can fetch in and out properties concurrently like https://stackoverflow.com/a/59664248

mcf-browser/src/back-end/utils.js
shifucun: Just updated the v1 API, please use the one Julia suggested.

cwcrystal8: Spoke to Bo and the new signature is live now, the next commit will have this change!

mcf-browser/src/back-end/utils.js
cwcrystal8: Added in next commit!

mcf-browser/src/back-end/utils.js
```diff
@@ -70,8 +70,8 @@ async function getRemotePropertyLabels(dcid) {
 async function getRemotePropertyValues(dcid, label, isInverse) {
   const direction = isInverse ? 'in' : 'out';
   const targetUrl =
-      (API_ROOT + '/v1/property/' + direction + '/' + label +
-      '/values/' + dcid);
+      (API_ROOT + '/v1/property/values/' + direction + '/' + dcid +
```
shifucun, line 73: nit: can use template literals for better readability: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

mcf-browser/src/back-end/utils.js
cwcrystal8: Changed in next commit!

mcf-browser/src/back-end/utils.js
cwcrystal8: Changed in next commit!

cwcrystal8:
Changes since last review:
- API calls changed to new signature
- Sequential API calls switched to simultaneous API calls
- URLs formed with template literals instead of string addition

shifucun (approved):
thanks for the update

PR #133, opened 2022-07-28
Fix MCF Browser File Matching

cwcrystal8:
This PR changes the way that files are parsed. Now the expected behavior is:

- Multiple MCF files can be uploaded with no error
- Multiple CSV files can be uploaded with no error
- Only one TMCF file will be used for all CSV files. If multiple are uploaded, the most recently uploaded TMCF file will be used and an error will be displayed to the user
- All files are reparsed every time a new file is loaded
- CSV files can be uploaded separately from the TMCF file


mcf-browser/src/back-end/server-api.ts
```diff
@@ -37,9 +37,32 @@ type ParseFileResponse = {
...
+
   const finalReturn: ParseFileResponse = {'errMsgs': [], 'localNodes': []};
 
+  // Find TMCF file, if it exists
+  let tmcfFile = null;
+  for (const file of fileList) {
+    const fileName = (file as File).name;
+    const fileExt = fileName.split('.').pop();
+
+    if(fileExt === "tmcf"){
```
chejennifer, line 51: nit: add space after if

mcf-browser/src/back-end/server-api.ts
```diff
@@ -37,9 +37,32 @@ type ParseFileResponse = {
...
+      if (tmcfFile) {
+        // If another TMCF file was found, throw an error
+        finalReturn['errMsgs'] = finalReturn['errMsgs'].concat([{
+          'file': fileName,
```
chejennifer, line 55: maybe instead of the current file that is going to be used, we can put the file that got skipped (tmcfFile)?

mcf-browser/src/back-end/utils.ts
```diff
@@ -33,6 +33,7 @@ const ERROR_MESSAGES =
       'assert-noCur': 'current node must be set before declaring properties',
       'parse-noColon': 'missing \':\', incorrect mcf triple format',
       'parse-noLabel': 'missing property label',
+      'multiple-tmcf': 'multiple TMCF files were uploaded, only the last one was used',
```
chejennifer, line 36: nit: we want constants to be in all caps and with underscores, could we update all of these

eg. `'multiple-tmcf'` => MULTIPLE_TMCF & to use it, we would do: ERROR_MESSAGES.MULTIPLE_TMCF

chejennifer (commented):
nice job!

mcf-browser/src/back-end/server-api.ts
cwcrystal8: changed in next commit!

mcf-browser/src/back-end/server-api.ts
cwcrystal8: Good catch! Updated in the next commit

mcf-browser/src/back-end/utils.ts
cwcrystal8: Changed!

cwcrystal8:
ptal, changes since last commit:

- Fix whitespace issues
- Change error to display the replaced file instead of the file that's replacing
- Change error keys to be all caps constants

chejennifer (approved):
thanks for the updates!

PR #164, opened 2022-12-17
Revert project set to fix continuous build error

shifucun:
This is added in https://github.com/datacommonsorg/tools/pull/162 and causes build error: https://groups.google.com/a/google.com/g/datacommons-alerts/c/dJeTKN-k1Kw/m/6dD9KNJFBAAJ

### DIFF
--- a/periodic_builds/build_all_repos.sh
+++ b/periodic_builds/build_all_repos.sh
@@ -51,6 +51,8 @@ function submit_cloud_build {
 	# if $FAILED_FOLDER
 	outfile="$1.out.$repo" # header_file + buildlog_file gets written to outfile
 
+	gcloud config set project datcom-ci
+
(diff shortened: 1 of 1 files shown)


Fructokinase (dismissed):
Thanks for the quick change! 

Took a brief look at the script. It seems like the periodic build builds individual repos. But I can't seem to find the build of each repo in Cloud Build. Did you intend to set the project id instead of account so that each individual repo build shows up on datcom-ci?

shifucun:
> Thanks for the quick change!
> 
> Took a brief look at the script. It seems like the periodic build builds individual repos. But I can't seem to find the build of each repo in Cloud Build. Did you intend to set the project id instead of account so that each individual repo build shows up on datcom-ci?

That's a good point, changed to project.

Fructokinase (approved):

PR #168, opened 2022-12-21
Remove manual scaling for BT automation

Fructokinase:
Use auto-scaling feature instead. 

- For custom DC, this completes the BT automation setup because Terraform configures autoscaling for BT.
- For main DC, this should prevent the problem of unintended scale down of BT during long running jobs from multiple import groups.

Using datcom-store-dev, I copied over a few [caches](https://pantheon.corp.google.com/storage/browser/automation_control_test/alex20221220/cache?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&mods=-monitoring_api_staging&project=google.com:datcom-store-dev&prefix=&forceOnObjectsSortingFiltering=false); 1 infrequent and a few frequents. Manually copied over a few init.txt files over to trigger concurrent Dataflow jobs. # of BT nodes scaled to the max(200) and [frequent finished](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2022-12-20_17_29_25-17789049836021027587;graphView=0?project=google.com:datcom-store-dev) in ~30min. [Infrequent is still running](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2022-12-20_17_05_38-14593584664668983817;graphView=0?project=google.com:datcom-store-dev). 

Also a few small improvements
- Add a hash as a part of the gcf source archive name to force an upload everyfile source changes.
- Delete BT tables when Dataflow fails to launch or GCS write fails.

### DIFF
--- a/bigtable_automation/gcf/custom_instance.go
+++ b/bigtable_automation/gcf/custom_instance.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"log"
 	"os"
-	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -29,8 +28,6 @@ func customInternal(ctx context.Context, e GCSEvent) error {
 	bucket := os.Getenv("bucket")
 	instance := os.Getenv("instance")
 	cluster := os.Getenv("cluster")
-	nodesHigh := os.Getenv("nodesHigh")
-	nodesLow := os.Getenv("nodesLow")
 	dataflowTemplate := os.Getenv("dataflowTemplate")
 	if projectID == "" {
 		return errors.New("projectID is not set in environment")
@@ -44,15 +41,6 @@ func customInternal(ctx context.Context, e GCSEvent) error {
 	if dataflowTemplate == "" {
 		return errors.New("dataflowTemplate is not set in environment")
 	}
-	// Get low and high nodes number
-	nodesH, err := strconv.Atoi(nodesHigh)
-	if err != nil {
-		return errors.Wrap(err, "Unable to parse 'nodesHigh' as an integer")
-	}
-	nodesL, err := strconv.Atoi(nodesLow)
-	if err != nil {
-		return errors.Wrap(err, "Unable to parse 'nodesLow' as an integer")
-	}
 	// Get table ID.
 	// e.Name should is like "**/<user>/<import>/control/<table_id>/launched.txt"
 	parts := strings.Split(e.Name, "/")
@@ -64,11 +52,6 @@ func customInternal(ctx context.Context, e GCSEvent) error {
 	}
 	tableID := parts[idxTable]
 	rootFolder := "gs://" + bucket + "/" + strings.Join(parts[0:idxControl], "/")
-
-	numNodes, err := getBTNodes(ctx, projectID, instance, cluster)
-	if err != nil {
-		return err
-	}
 
 	if strings.HasSuffix(e.Name, initFile) {
 		log.Printf("[%s] State Init", e.Name)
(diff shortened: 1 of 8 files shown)


bigtable_automation/gcf/custom_instance.go
```diff
@@ -29,8 +28,6 @@ func customInternal(ctx context.Context, e GCSEvent) error {
 	bucket := os.Getenv("bucket")
 	instance := os.Getenv("instance")
 	cluster := os.Getenv("cluster")
-	nodesHigh := os.Getenv("nodesHigh")
```
Spaceenter, line 32: Also remove "nodesHigh" and "nodesLow" from yaml files? See: https://github.com/datacommonsorg/tools/search?q=nodesHigh

Fructokinase: Removed for local and prod. yamls under private still references nodes high/low for deploy.sh, so I'll leave them there for now. Thanks!

pradh (commented):
Nice improvements.  Any thoughts on why it doesn't scale past 200?

Fructokinase:
> Nice improvements. Any thoughts on why it doesn't scale past 200?

That's because the BT on datcom-store-dev is set to 200 by me. I can't seem to set any higher due to quotas. Do quotas expire?

pradh:
> > Nice improvements. Any thoughts on why it doesn't scale past 200?
> 
> That's because the BT on datcom-store-dev is set to 200 by me. I can't seem to set any higher due to quotas. Do quotas expire?
> 
> Update: quotas are per zone and I found the zone with 330 quota, may need to test again after this with 300.
> 
> Update 2: With 300 nodes, infrequent finished in 2hrs56min, roughly the same as the last infrequent run on datcom-store (2hr55min, same data). Downside of autoscaling is that it takes 1hr30min for 300 nodes to come back down to 30 (seems like probing happens every 5min or so to scale down bit by bit).
> 

That time is pretty good!

bigtable_automation/gcf/custom_instance.go
```diff
@@ -84,35 +67,24 @@ func customInternal(ctx context.Context, e GCSEvent) error {
...
-		// Called when the state-machine moves to Completed state from Launched.
-		if numNodes == nodesH {
-			// Only scale down BT nodes when the current high node is set up this config.
-			// This requires different high nodes in different config.
-			if err := scaleBT(ctx, projectID, instance, cluster, int32(nodesL)); err != nil {
-				return err
-			}
-		}
-		// TODO: else, notify Mixer to load the BT table.
-		log.Printf("[%s] Completed work", e.Name)
```
pradh, line 115: Can you restore the completed work message?

bigtable_automation/gcf/prod.go
```diff
@@ -99,36 +82,22 @@ func prodInternal(ctx context.Context, e GCSEvent) error {
...
-		// Called when the state-machine moves to Completed state from Launched.
-		if numNodes == nodesH {
-			// Only scale down BT nodes when the current high node is set up this config.
-			// This requires different high nodes in different config.
-			if err := scaleBT(ctx, projectID, instance, cluster, int32(nodesL)); err != nil {
-				return err
-			}
-		}
-		// TODO: else, notify Mixer to load the BT table.
-		log.Printf("[%s] Completed work", e.Name)
```
pradh, line 131: ditto

pradh (dismissed):

bigtable_automation/gcf/custom_instance.go
Fructokinase: done.

bigtable_automation/gcf/prod.go
Fructokinase: done.

pradh (approved):

PR #169, opened 2022-12-29
Fix prod BT gcf

Fructokinase:
This is a fix for #168. Accidentally deleted the part that creates a table for prod gcf.

The fix was already deployed to prod gcf, this code change is just a follow up.


juliawu (approved):
Thanks for the fix!

PR #186, opened 2023-02-10
[Custom DC] Add a config generator

Fructokinase:
Config generator generates a config based on the environment variable "dataDirectory".


bigtable_automation/gcf/custom_dc/local.yaml
```diff
@@ -1,7 +1,8 @@
...
+projectID: "datcom-mixer-encode"
 instance: "prophet-test"
 cluster: "prophet-test-c1"
 dataflowTemplate: "gs://datcom-dataflow-templates/templates/flex/csv_to_bt.json"
 tempLocation: "gs://datcom-store-dev-resources/dataflow/tmp"
-controllerTriggerTopic: "projects/datcom-204919/topics/custom-import-notification-test"
-bucket: "automation_control_test"
+controllerTriggerTopic: "projects/datcom-204919/topics/private-import-notification-dev"
+bucket: "datcom-mixer-encode-resources"
+dataDirectory: "demo/data/"
```
shifucun, line 8: the data directory "demo" is set by the user, this may not work if there are other folders?

bigtable_automation/gcf/gcs.go
```diff
@@ -0,0 +1,191 @@
...
+// pathToDataFolder is the "gs://"-prefixed folder that contains raw data.
+// For the folder structure that is expected, please see,
+// https://docs.datacommons.org/custom_dc/upload_data.html
+func GenerateManifest(ctx context.Context, bucket, pathToDataFolder string) (*pb.DataCommonsManifest, error) {
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	importGroupName := "custom"
```
shifucun, line 107: can we infer the import group name from trigger? As the folder structure is fixed

bigtable_automation/gcf/manifest.go
```diff
@@ -0,0 +1,104 @@
+package gcf
+
+import (
+	pb "github.com/datacommonsorg/tools/bigtable_automation/gcf/proto"
+	"google.golang.org/protobuf/proto"
+)
+
+type ImportParams struct {
```
shifucun, line 8: Why not using the protobuf struct directly but have go struct here and do conversions?

shifucun (approved):
Have some high level questions, but can submit now and come back for them later.

PR #190, opened 2023-02-10
[Custom DC] Determine root import directory on the fly and use that as import group name

Fructokinase:


bigtable_automation/gcf/gcs.go
```diff
@@ -104,7 +104,12 @@ func GenerateManifest(ctx context.Context, bucket, pathToDataFolder string) (*pb
 		return nil, err
 	}
 
-	importGroupName := "custom"
+	// IMPORTANT NOTE: importGroupName has a character limit of 21
```
shifucun, line 107: why has this limit?

bigtable_automation/gcf/controller_trigger.go
```diff
@@ -102,3 +103,30 @@ func (s CustomDCPubSubMsg) Publish(ctx context.Context, p PublishConfig) error {
...
+// Find the root directory for data imports on gcs.
+// For data import directory structures, please refer to
+// https://docs.datacommons.org/custom_dc/upload_data.html
+//
+// Root directory in the above example is "gcs_folder"
+// Example:
+// findRootImportDirectory("/somepath/gcs_folder/internal/control/trigger.txt")
+// returns "/somepath/gcs_folder/"
+//
+// Note manifest generator requires the "/" at the end.
```
shifucun, line 116: Maybe start to add test, can use mock GCS as suggested in https://stackoverflow.com/questions/39814108/how-to-unit-test-google-cloud-storage-in-golang

Fructokinase: Will leave this as a todo

bigtable_automation/gcf/gcs.go
Fructokinase: > why has this limit?

This is our internal limit

shifucun (approved):

PR #192, opened 2023-02-13
Move gcf functions out of bigtable_automation

shifucun:
As there are several gcf now, and they are not all related to bigtable automation


Fructokinase (approved):

pradh:
Can we call the top-level directory something other than "gcf"?  :)

PR #195, opened 2023-02-14
Sync all references of BT loader templates

Fructokinase:
Recent changes in dataflow templates include switching from classic to flex, and versioning it (so at any time we can switch to old ones in case new ones break). This resulted in some inconsistencies of which templates are used.

- Current latest version of template is gs://gs://datcom-templates/templates/flex/csv_to_bt_0.0.3.json
- All templates of testing purposes are stored in gs://datcom-dataflow-templates/templates


bigtable_automation/gcf/custom_dc/mixer-autopush.yaml
```diff
@@ -3,5 +3,5 @@ instance: "dc-graph"
 cluster: "dc-graph-c1"
 nodesHigh: "3"
 nodesLow: "1"
-dataflowTemplate: "gs://datcom-templates/templates/csv_to_bt"
+dataflowTemplate: "gs://gs://datcom-templates/templates/flex/csv_to_bt_0.0.3.json"
```
shifucun, line 6: duplicate gs://

shifucun (approved):

bigtable_automation/gcf/custom_dc/mixer-autopush.yaml
Fructokinase: fixed.

PR #196, opened 2023-02-14
[GCF] Re-organize folder structure

shifucun:
* Move cloud functions entry point functions to the root folder, otherwise they can not be deployed.
* Use /gcp and /local sub folder for each cloud function, with /config under /gcp
* Updated deployment script with more check
* Add .gitignore
* Tested with local environment and GCP deployment (for branch cache GCF)


Fructokinase (approved):

shifucun:
thanks!

PR #198, opened 2023-02-14
[Custom DC] Change import file finder logic

Fructokinase:
Currently files for data import are found by iteratively finding folders and files in gcs. This causes a bug where folders (determined by "/" ending) is not consistently found due to unknown reasons. 

This PR finds all paths with a prefix (root directory of import, also the import group name), and uses a ImportGroupFiles struct to pass data into manifest proto.


gcf/lib/path.go
```diff
@@ -0,0 +1,110 @@
...
+// CollectImportFiles takes a list of gcs file paths and construct ImportGroupFiles.
+func CollectImportFiles(paths []string) (*ImportGroupFiles, error) {
+
+	dataSource2DatasetNames := map[string][]string{}
+	datasetName2DataFiles := map[string]DataFiles{}
+
+	if len(paths) == 0 {
+		return nil, errors.New("No path found in import group files")
+	}
+	importGroupName := strings.Split(paths[0], "/")[0]
```
shifucun, line 47: Add a function comment that all the paths here are under the same import group folder.

gcf/lib/path.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	// IMPORTANT NOTE: importGroupName has a character limit of 21
+	importGroupName := strings.Split(paths[0], "/")[0]
+	if len(importGroupName) > 20 {
+		importGroupName = importGroupName[:20]
+	}
+
+	for _, path := range paths {
+		// Path is expected to be like the following:
+		// 	"demo/data/source2/solar/output.csv"
```
shifucun, line 54: line54 missing a subfolder?

gcf/lib/path.go
```diff
@@ -0,0 +1,116 @@
...
+		// 	"demo/data/source1/smokepm/subfolder/data.tmcf"
+		folderList := strings.SplitN(path, "/", 5)
+		if folderList[0] != importGroupName {
```
shifucun, line 57: This is comparing with the truncated version. Should the truncation happen in the end?

gcf/lib/path.go
```diff
@@ -0,0 +1,116 @@
...
+
+// DataFiles represent a logical unit of files for a single dataset.
+type DataFiles struct {
+	TMCFPath string
+	CSVPaths []string
+}
+
+// ImportGroupFiles represent the collection of paths for a single import group.
+type ImportGroupFiles struct {
+	DataSource2DatasetNames map[string][]string
```
shifucun, line 32: nit: no need to have "data" in all the variable names. Can just be "Source2Dataset", "Dataset2Files". It makes the code below easier to read.

gcf/custom/manifest.go
```diff
@@ -22,129 +22,109 @@ import (
...
+		}
+		// skip folders
+		if strings.HasSuffix(attrs.Name, "/") {
+			continue
+		}
+		log.Printf("FindFiles in %s got %s", pathToDataFolder, attrs.Name)
+		paths = append(paths, attrs.Name)
+	}
+
+	// IMPORTANT NOTE: importGroupName has a character limit of 21
```
shifucun, line 118: importGroupName is not used here?

gcf/lib/path.go
Fructokinase: That is intentional, <dataset>/a.tmcf and <dataset>/subfolder/a.tmcf are both valid.

gcf/lib/path.go
Fructokinase: done.

gcf/lib/path.go
Fructokinase: Thanks for the catch, fixed and added a test case.

gcf/lib/path.go
Fructokinase: done.

gcf/custom/manifest.go
Fructokinase: bad merge, fixed.

gcf/lib/path.go
shifucun: oh, actually line 52 "subfolder" is not supported?

shifucun (dismissed):

gcf/lib/path.go
Fructokinase: added a constraint.

PR #199, opened 2023-02-15
Set a global import group instance id in pubsub message

shifucun:
This will be used by controller runner as the instance_id and will deprecate "import_name" field.

Also made some minor formatting change for long line arguments


Fructokinase (approved):

shifucun:
thanks!

PR #207, opened 2023-02-21
[Custom DC] Fix terraform deployment for BT automation

Fructokinase:
Update to last week's refactoring


shifucun (dismissed):
I guess an e2e test would capture issue like this?

Fructokinase:
> I guess an e2e test would capture issue like this?

Yes. with this PR, I confirmed that all resources in BT automation Terraform are deletable with terraform destroy.

PR #208, opened 2023-02-21
parse provenance.json in layout generation

shifucun:
Update deployment projects too.

### DIFF
--- a/cloudbuild.deploy.yaml
+++ b/cloudbuild.deploy.yaml
@@ -20,5 +20,5 @@ steps:
       - |
         set -e
         ./gcf/custom/gcp/deploy.sh datcom-stanford
-        ./gcf/custom/gcp/deploy.sh datcom-magiceye-dev
+        ./gcf/custom/gcp/deploy.sh datcom-mixer-encode
         ./gcf/custom/gcp/deploy.sh custom-dc-test
--- a/gcf/custom/layout.go
+++ b/gcf/custom/layout.go
@@ -20,6 +20,8 @@ import (
 	"strings"
 )
 
+const provFile = "provenance.json"
+
 type Table struct {
 	tmcf string
 	csv  []string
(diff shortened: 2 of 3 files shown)


gcf/custom/layout.go
```diff
@@ -62,8 +66,10 @@ func BuildLayout(root string, objects []string) (*Layout, error) {
 	}
 	// Put files under each Import
 	for _, parts := range files {
-		if len(parts) < 2 {
-			logging("Ignore file", parts)
+		if len(parts) == 1 {
+			if parts[0] == provFile {
+				layout.prov = provFile
+			}
 			continue
```
Fructokinase, line 73: nit: Can we keep the log please? It would be helpful to know if a file was ignored.

cloudbuild.deploy.yaml
```diff
@@ -20,5 +20,5 @@ steps:
       - |
         set -e
         ./gcf/custom/gcp/deploy.sh datcom-stanford
-        ./gcf/custom/gcp/deploy.sh datcom-magiceye-dev
+        ./gcf/custom/gcp/deploy.sh datcom-mixer-encode
```
Fructokinase, line 23: For my own knowledge, settings here are not auto-deployed right? I think datcom-mixer-encode is actively used for climate trace instance?

Fructokinase (approved):

gcf/custom/layout.go
shifucun: Good catch, not meant to remove those.

cloudbuild.deploy.yaml
shifucun: These will be auto-deployed. I think it's okay since this only affects GCF (not any user visible parts), and we almost always want to deploy latest to these instances.

Fructokinase (approved):

PR #215, opened 2023-03-01
[Custom DC] Disable the creation of BigTable instance by default

Fructokinase:
This is the most costly component of Custom DC.  The BT instance can be create right before the first import instead by running through the Terraform script again with -var=enable_bigtable_instance=true.


shifucun (approved):
This is nice. Can you add instruction for this at the beginning of this page https://docs.datacommons.org/custom_dc/upload_data.html

PR #216, opened 2023-03-01
Support provenance in import level

shifucun:
Import group (source) level provenance.json is used first. If import level provenance.json exists, then prefer it.

### DIFF
--- a/gcf/custom/manifest.go
+++ b/gcf/custom/manifest.go
@@ -48,6 +48,58 @@ func computeDataMCF(bucket, root, im, tab string, mcf []string) []string {
 	return bigstoreMCFs
 }
 
+func computeSource(
+	ctx context.Context, reader Reader, bucket, root, provFile string,
+) (
+	*pb.DataCommonsManifest_DatasetSource, error,
+) {
+	// Populate data source that should be specified at import group level.
+	if provFile != "" {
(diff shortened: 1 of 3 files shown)


Fructokinase (dismissed):

gcf/custom/manifest.go
```diff
@@ -48,6 +48,58 @@ func computeDataMCF(bucket, root, im, tab string, mcf []string) []string {
...
 
+func computeSource(
+	ctx context.Context, reader Reader, bucket, root, provFile string,
+) (
+	*pb.DataCommonsManifest_DatasetSource, error,
+) {
+	// Populate data source that should be specified at import group level.
+	if provFile != "" {
+		provJSON, err := reader.ReadObject(
+			ctx, bucket, path.Join(root, "data", provFile))
```
Fructokinase, line 59: As per [doc](https://docs.datacommons.org/custom_dc/upload_data.html), should this be path.Join(root, provFile)?

Fructokinase (changes requested):

gcf/custom/manifest.go
shifucun: good catch, the doc needs to be updated, "data" folders should have all the user data

Fructokinase (approved):

PR #221, opened 2023-06-01
Add gcf config for feeding-america and fix a bug of GCS write error

shifucun:


gcf/custom/manifest.go
```diff
@@ -193,10 +194,17 @@ func ComputeManifest(
 					manifestImport.McfProtoUrl,
 					filepath.Join("/bigstore", bucket, root, "data", im, tab, "graph.tfrecord@*.gz"),
 				)
+				manifestImport.McfUrl = append(
+					manifestImport.McfUrl,
+					filepath.Join("/bigstore", bucket, root, "data", im, tab, "graph.tfrecord@*.gz"),
```
Spaceenter, line 199: mcf_url should not be graph.tfrecord
mcf_url should be something like   ..../.../nodes.mcf@N

Spaceenter (changes requested):

gcf/custom/manifest.go
shifucun: I don't see nodes.mcf@N created in the GCS though. Will they be created if I add this?

Spaceenter (approved):

gcf/lib/common.go
```diff
@@ -86,9 +86,17 @@ func WriteToGCS(ctx context.Context, path, data string) error {
 		return errors.Wrap(err, "Failed to create gcsClient")
 	}
 	w := gcsClient.Bucket(bucket).Object(object).NewWriter(ctx)
-	defer w.Close()
-	_, err = fmt.Fprint(w, data)
-	return errors.WithMessagef(err, "Failed to write data to %s/%s", bucket, object)
+	_, writeErr := fmt.Fprint(w, data)
+	closeErr := w.Close()
```
Spaceenter, line 90: Move this to after line 94?

Spaceenter (approved):

gcf/lib/common.go
shifucun: maybe need to close here even though write has error?

PR #227, opened 2023-09-15
Fix cron job script issue

shifucun:
Tested by run the script locally to trigger two tests, one with substitutions and one without.


keyurva (approved):
Thanks Bo. This will be very useful.

shifucun:
Thanks for review!

PR #237, opened 2023-09-27
Add skipNamesInSvgCache env to custom DC cloud functions

shifucun:


gcf/custom/handler.go
```diff
@@ -86,6 +90,7 @@ func HandleTriggerFlow(ctx context.Context, bucket, root string) error {
 		"bigstore_cache_directory":   bigstoreCacheDirectory,
 		"bigstore_control_directory": bigstoreControlDirectory,
 		"produce_sv_and_svg_mcfs":    "true",
+		"skip_names_in_svg_cache":    skipNamesInSvgCache,
```
pradh, line 93: Oh, I think just setting any value to it enables it per logic in https://source.corp.google.com/piper///depot/google3/datacommons/import/controller/runner/controller_runner.cc;l=81-89.

So I think set the value in attributes only if its true?

pradh (approved):

gcf/custom/handler.go
shifucun: Done

pradh (approved):
Thanks

PR #243, opened 2023-10-17
Only run screenshot and sanity check for daily cron job

shifucun:
Adversarial tests now runs for 1.5 hours. Turn that off before we can make it faster.

### DIFF
--- a/periodic_builds/builds.txt
+++ b/periodic_builds/builds.txt
@@ -5,5 +5,6 @@ mixer/build/ci/cloudbuild.test.yaml
 website/build/ci/cloudbuild.npm.yaml
 website/build/ci/cloudbuild.py.yaml
 website/build/ci/cloudbuild.webdriver.yaml
-website/build/ci/cloudbuild.website_testing.yaml _DOMAIN=dev.datacommons.org,_TEST=all
+website/build/ci/cloudbuild.website_testing.yaml _DOMAIN=dev.datacommons.org,_TEST=screenshot
+website/build/ci/cloudbuild.website_testing.yaml _DOMAIN=dev.datacommons.org,_TEST=sanity
 api-python/cloudbuild.yaml


keyurva:
@pradh - FYI. Adversarial is being removed from daily runs since Bo found that the tests are taking much longer of late (from 30-40 mins to 1:30).

shifucun:
> @pradh - FYI. Adversarial is being removed from daily runs since Bo found that the tests are taking much longer of late (from 30-40 mins to 1:30).

One run from today: https://pantheon.corp.google.com/cloud-build/builds;region=global/6d80818e-70be-47bb-96b9-cfed7fe9ef23;step=3?mods=-monitoring_api_staging&project=datcom-ci. Takes 1 hour 15min.

pradh:
> @pradh - FYI. Adversarial is being removed from daily runs since Bo found that the tests are taking much longer of late (from 30-40 mins to 1:30).

Do we know if the delay is from client side issues vs. latency per query?

keyurva:
> > @pradh - FYI. Adversarial is being removed from daily runs since Bo found that the tests are taking much longer of late (from 30-40 mins to 1:30).
> 
> Do we know if the delay is from client side issues vs. latency per query?

We don't have data so can't say for sure. The script however hasn't changed in a while. Have we seen any latency changes in the detect-and-fulfill endpoint of late?

pradh:
> > > @pradh - FYI. Adversarial is being removed from daily runs since Bo found that the tests are taking much longer of late (from 30-40 mins to 1:30).
> > 
> > 
> > Do we know if the delay is from client side issues vs. latency per query?
> 
> We don't have data so can't say for sure. The script however hasn't changed in a while. Have we seen any latency changes in the detect-and-fulfill endpoint of late?

Latency is unfortunately also a function of the data changes.

I think we can also mine the logs for the latencies.

Might consider one of these tests to log timing info so we can detect regressions more easily and answer questions like these.  Is the sanity checker a candidate for that?

keyurva:
> > > > @pradh - FYI. Adversarial is being removed from daily runs since Bo found that the tests are taking much longer of late (from 30-40 mins to 1:30).
> > > 
> > > 
> > > Do we know if the delay is from client side issues vs. latency per query?
> > 
> > 
> > We don't have data so can't say for sure. The script however hasn't changed in a while. Have we seen any latency changes in the detect-and-fulfill endpoint of late?
> 
> Latency is unfortunately also a function of the data changes.
> 
> I think we can also mine the logs for the latencies.
> 
> Might consider one of these tests to log timing info so we can detect regressions more easily and answer questions like these. Is the sanity checker a candidate for that?

That's a good question. The current sanity checker is FE and web driver focussed. We can either augment it for API sanity or it may be cleaner to create a separate checker for it.

PR #252, opened 2024-02-09
change sdk to 2.43.0 with BigtableOptions and set maxNumWorkers to 1000

n-h-diaz:
This is because the latest SDK (2.53.0) was causing the infrequent load to fail (leading to increased latency and other production issues). Limiting the workers is to help manage the CPU usage, which spikes during load jobs.

### DIFF
--- a/bigtable_automation/java/dataflow/csv2bt-template-metadata.json
+++ b/bigtable_automation/java/dataflow/csv2bt-template-metadata.json
@@ -40,6 +40,13 @@
             "isOptional": true
         },
         {
+            "name": "bigtableMaxNumWorkers",
+            "label": "Bigtable maximum number of workers",
+            "helpText": "The maximum number of workers for Dataflow jobs.",
+            "paramType": "TEXT",
+            "isOptional": true
+        },
+        {
             "name": "tempLocation",
             "label": "GCP temp location",
             "helpText": "A GCS bucket where the running jobs will have access to.",
--- a/bigtable_automation/java/dataflow/csv2bt-template-metadata.json
+++ b/bigtable_automation/java/dataflow/csv2bt-template-metadata.json
@@ -40,8 +40,8 @@
             "isOptional": true
         },
         {
-            "name": "bigtableMaxNumWorkers",
-            "label": "Bigtable maximum number of workers",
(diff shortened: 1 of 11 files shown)


bigtable_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CloudBigtableOptions.java
```diff
@@ -28,4 +28,10 @@ public interface CloudBigtableOptions extends DataflowPipelineOptions {
 
   @SuppressWarnings("unused")
   void setBigtableAppProfileId(ValueProvider<String> bigtableAppProfileId);
+
+  @Description("The maximum number of workers for Dataflow jobs.")
+  ValueProvider<String> getBigtableMaxNumWorkers();
```
keyurva, line 33: Is there a way to specify defaults?

bigtable_automation/java/dataflow/csv2bt-template-metadata.json
```diff
@@ -39,6 +39,13 @@
             "paramType": "TEXT",
             "isOptional": true
         },
+        {
+            "name": "bigtableMaxNumWorkers",
+            "label": "Bigtable maximum number of workers",
```
pradh, line 44: nit:  dataflowMaxNumWorkers (the variable isn't BT related right?)

but if its churn, feel free to ignore too

pradh (approved):
Nicely done!

pradh:
Could you please add context to the description on why the change is being made?

bigtable_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CloudBigtableOptions.java
n-h-diaz: I think the default behavior in this case is to use the Dataflow autoscaling

bigtable_automation/java/dataflow/csv2bt-template-metadata.json
n-h-diaz: done

pradh (approved):
Thanks for the change!

PR #272, opened 2024-08-20
Push LLM updates to git repo

pradh:
Changes:

* Modify RIG in-context flow to use a base LLM (eg Gemma 9B) and then an annotator LLM (eg, Gemini 1.5 Pro), instead of a single LLM call that answers + annotates.  This has been tested with eval runs internally.
* Add OpenAI LLM impl.  This has been tested internally.
* Add HF Basic and Pipeline API impls.

Colab NB with HF API examples: https://colab.research.google.com/drive/18bdz4VXhSBsXskv696H7G-zHn7vrqGaE?resourcekey=0-LA7cfsahChB3f2E6vUcujQ#scrollTo=4qqZvnRGRG-X

### DIFF
--- a/llm/data_gemma/__init__.py
+++ b/llm/data_gemma/__init__.py
@@ -25,6 +25,7 @@ from data_gemma import rig
 LLM = base.LLM
 LLMCall = base.LLMCall
 GoogleAIStudio = google_api.GoogleAIStudio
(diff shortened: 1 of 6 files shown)


chejennifer (approved):
thanks for adding!

PR #276, opened 2024-08-29
Strip trailing non-alphanumeric characters from importGroup when creating config.textproto

n-h-diaz:


gcf/custom/manifest.go
```diff
@@ -25,6 +26,40 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+func isAlphaNumeric(c byte) bool {
+	return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')
+}
+
+func computeImportGroup(root string) (string, error) {
```
shifucun, line 33: getImportGroupName?

gcf/custom/manifest.go
```diff
@@ -25,6 +26,40 @@ import (
...
+
+	// Strip any trailing non-alphanumeric characters.
+	i := len(importGroup)
+	for !isAlphaNumeric(importGroup[i-1]) {
+		i--
+		if i == 0 {
+			break
+		}
+	}
+	importGroup = importGroup[:i]
```
shifucun, line 54: why not do the trimming first then take the first 20?

gcf/custom/manifest.go
n-h-diaz: Done

gcf/custom/manifest.go
n-h-diaz: This is to handle cases where the truncated name ends with symbols 
(e.g. "sustainable_systems_lab" --> "sustainable_systems_" --> "sustainable_systems") 

Alternatively I could just remove all non-alphanumeric characters from the name

Wdyt?

shifucun: Yea, that sg and leaves more room for truncation!

shifucun: oh, do you want to make the change to remove non-alphanum chars?

n-h-diaz: Done

shifucun (approved):
thanks!

PR #349, opened 2026-05-14
Implement open-source maintainer playbook

beets:
Implementing the suggestions from the playbook. LMK if there are updates we should make.


dwnoble (approved):
Thank you @beets !

PR #351, opened 2026-05-28
Setup Data Weaver Project Structure w/ Basic TLDraw Scaffold

PauloMFJ:


dataweaver/apps/web/src/app/not-found.tsx
```diff
@@ -0,0 +1,3 @@
+import { Page404 } from '~/components/scopes/page-404';
```
nick-nlb: The [Google TS Style Guide](https://google.github.io/styleguide/tsguide.html) requires `camel_case` file names (see [5.2.4](https://google.github.io/styleguide/tsguide.html#identifiers-imports)).

`not-found.tsx` to `not_found.tsx` (applies to other files as well).

In the `CONTRIBUTING.md` file at the root of the `tools` repo, there is a link to all the style guides (although in practice the above TS one will be the primary one used in this repo).

dataweaver/apps/web/src/components/primitives/link.tsx
```diff
@@ -0,0 +1,52 @@
+import type { LinkProps } from 'next/link';
+import NextLink from 'next/link';
+import type { ComponentPropsWithRef, Ref } from 'react';
+
+type Props = {
+	ref?: Ref<HTMLAnchorElement>;
```
nick-nlb: While this isn't explicitly stated in the style guide, the "spacing/tab/number of spaces" guideline is to use spaces over tabs, with two spaces per indent.

dataweaver/packages/tokens/src/colors.json
```diff
@@ -0,0 +1,9 @@
+{
+	"black": [0, 0, 0],
```
nick-nlb: For design tokens, colors should be semantic rather than to refer to explicit colors, so that the tokens can be modified to very different themes. For example, `color-white` could be `color-canvas-background`, etc.

dataweaver/apps/web/src/styles/tokens/_generated.module.scss
```diff
@@ -0,0 +1,21 @@
+// ⚠️ AUTO-GENERATED — Do not edit. Modify the JSON tokens in 'packages/tokens/src/' and run 'pnpm generate:tokens'
+
+// Variables
+$breakpoint-tablet: 768;
+$breakpoint-laptop: 1280;
+$breakpoint-desktop: 1600;
+
+// Colors
+$color-black: rgb(0 0 0);
```
nick-nlb: How do we envision themability of the design tokens here?

If we take something like "color-canvas-background": it looks like the process currently is:

- the generator takes that color and puts it into $color-canvas-background in this file.
- subsequent modules like tldraw.module.scss import that color directly
- a clone of this repo would have to update the core color.json file and regenerate to update the color (and in doing so, change files directly in the repo.

What do you think of a layer in between, where rather than the generator directly creating `$color-canvas-background: rgb(x y z);`, we have the generator create a set of CSS variables

```
:root {
  --color-canvas-background: x y z;
}

$color-canvas-background: rgb(var(--color-canvas-background));
```

A custom instance could override this by adding in CSS in a specified spot:
```
html:root {
  --color-canvas-background: a b c;
}

File: dataweaver/apps/web/src/styles/tokens/_generated.module.scss [hunk 5]
```diff
@@ -0,0 +1,21 @@
...
+$color-grey-almost-black: rgb(32 33 36);
+$color-grey-muted: rgb(68 71 70);
+$color-grey-light: rgb(227 231 237);
+$color-white: rgb(255 255 255);
+$color-blue: rgb(11 87 208);
+$color-blue-light: rgb(247 252 255);
```
nick-nlb: As a follow-up to the question/comment about themability:

We could consider dropping the SCSS variables completely, and just have the generator build CSS variables here (and then use those directly in the CSS modules).

dataweaver/apps/web/src/styles/tokens/generated.ts
```diff
@@ -0,0 +1,22 @@
+// ⚠️ AUTO-GENERATED — Do not edit. Modify the JSON tokens in 'packages/tokens/src/' and run 'pnpm generate:tokens'
```
nick-nlb: This file is causing a lint error (i.e. an error on running `npm run lint`). It's fixed by `npm run fix`, but the generator can be updated to output in a style the linter is happy with.

dataweaver/apps/web/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Next.js
+.next/
+next-env.d.ts
+
+# Generated
+src/server/storyblok/generated/globals.json
```
nick-nlb: Stray entry? (storyblok isn't referred to elsewhere)

dataweaver/README.md
```diff
@@ -1 +1,65 @@
-# Data Weaver Experiment
+# dataweaver
```
nick-nlb: nit: Data Weaver styled as two words.

dataweaver/README.md
```diff
@@ -1 +1,65 @@
...
+```
+apps/
+  web/         Next.js app (App Router)
+packages/
+  tokens/      Design tokens (JSON → generated .scss / .ts)
+```
+
+## Installation
+
+All commands run from the root of the monorepo.
```
nick-nlb: While as a standalone application this is a monorepo, it is really a sub-monorepo within a monorepo! (i.e. it is sitting inside the `tools` monorepo). This makes this line ambiguous, as it can be intepreted to refer back to the primary monorepo. Can we change it to "root of the `dataweaver` directory"?

dataweaver/README.md
```diff
@@ -1 +1,65 @@
...
+### Node.js version
+
+Install [nvm](https://github.com/nvm-sh/nvm) and run:
+
+```bash
+nvm use
+```
+
+### Dependencies
+
```
nick-nlb, line 27: The project looks like it is intended to be used with corepack.  If that is the intention, could we add instructions to run `corepack enable` above `pnpm i`.

dataweaver/biome.json
```diff
@@ -0,0 +1,91 @@
...
+	"formatter": {
+		"enabled": true,
+		"indentStyle": "tab"
+	},
+
+	"linter": {
+		"enabled": true,
+		"rules": {
+			"recommended": true,
+			"suspicious": {
```
nick-nlb: Let's elevate `no-explicit-any` to error (by default, it is being treated as a warning, but if `any` truly can't be avoided, the lint error should be suppressed with a comment rather than allowed through with a warning).

Likewise `use-const` should be an error, rather than a warning, as well as `no-unused-expression`

dataweaver/apps/web/src/components/primitives/link.tsx
```diff
@@ -0,0 +1,52 @@
+import type { LinkProps } from 'next/link';
+import NextLink from 'next/link';
+import type { ComponentPropsWithRef, Ref } from 'react';
+
+type Props = {
```
nick-nlb: When declaring shapes for objects, can we use interfaces over types?

(and to match with our historic convention, when declaring interfaces for components, ComponentNameProps, i.e. `interface LinkProps`).

The general preference for interfaces over types is discussed in 6.4 in the guide linked in an earlier comment.

dataweaver/biome.json
```diff
@@ -0,0 +1,91 @@
...
+			},
+			"complexity": {
+				"useOptionalChain": "off"
+			},
+			"style": {
+				"noDefaultExport": "error",
+				"useImportType": "error",
+				"useConsistentTypeDefinitions": {
+					"level": "error",
+					"options": { "style": "type" }
```
nick-nlb: If I'm interpreting this rule correctly, it might conflict with the guideline to use interfaces over types where possible (if so we would have to remove this).

dataweaver/apps/web/package.json
```diff
@@ -0,0 +1,27 @@
+{
+	"name": "@app/web",
+	"type": "module",
+	"scripts": {
+		"dev": "next dev",
+		"build": "next build",
+		"preview": "next start",
+		"lint:type-check": "tsc --noEmit",
+		"generate:tokens": "pnpm --filter @package/tokens generate -- \"../../apps/web/src/styles/tokens\""
```
nick-nlb: A bit of further reflection on the token generation in general:

When we are generating tokens here, we're generating them from the package directly into the application (i.e., the script directly outputs and writes the tokens, etc, from the token package to the app). 

An approach that would provide clearer isolation between the package and the app would be to have the token package generator write those tokens within the package itself (i.e. in a `dist` or `source`) and export those for availability in any consuming applications, in our case just the one. 

They could then be consumed by the app (and potentially any other applications that might ever use that package) . So the generator would generate (within the package) files that contains the CSS variables, any required SCSS and TS exports.

We would then import the css from the package (something like `import @package/tokens/theme.css` etc).

dataweaver/biome.json
```diff
@@ -0,0 +1,91 @@
...
+		"indentStyle": "tab"
+	},
+
+	"linter": {
+		"enabled": true,
+		"rules": {
+			"recommended": true,
+			"suspicious": {
+				"noConfusingVoidType": "off",
+				"noArrayIndexKey": "off",
```
nick-nlb: Some other linter rules that will help align with the guidelines (some of these might have slightly different names in Biome): `eqeqeq` (i.e. require identity over equality), `no-var`, `interface-over-type-literal` (see another comment I put relating to this).

nick-nlb (changes requested):
Thank you for all this setup and initial scaffolding!

dataweaver/apps/web/src/app/not-found.tsx
PauloMFJ: In our case I was following Next guidelines. i.e. `not_found` doesn't work as a Nextjs app route. We could:
1. Use `dash-case` for all app/ route files as needed.
2. And use `camel_case` for every other file.

Personally prefer to keep them the same across the app so just `dash-case` everywhere. But let me know your thoughts here.

dataweaver/apps/web/src/components/primitives/link.tsx
PauloMFJ: Sounds good :)

dataweaver/apps/web/src/components/primitives/link.tsx
PauloMFJ: Sure thing.

dataweaver/packages/tokens/src/colors.json
PauloMFJ: Personally prefer having a clean set of colors that we use across the app vs different values everywhere via specific naming. Find this pattern generally leads to more consistency across the app. In your example if for example neither `color-white` or `color-blue-light` work for canvas background even though color is visually very close to those I'd generally push for using one of the supported values over introducing a new token value. Thoughts?

dataweaver/apps/web/src/styles/tokens/_generated.module.scss
PauloMFJ: Related to my comment above. I suppose question here would be if theming is a requirement? I.e. will we ever need different apps with different --color-canvas-background? Or can we keep the single clean colour set that is unrelated to specific components.

dataweaver/apps/web/.gitignore
PauloMFJ: Good spot! Was from ref I ported some of the boilerplate from.

dataweaver/README.md
PauloMFJ: Good point - updated :)

dataweaver/biome.json
PauloMFJ: This relates to this comment + all other linting comments: I got AI to cross check our Biome / Stylint rule sets and match as close to the linked Google Styleguide. Let me know if anything is still not as you'd expect.

dataweaver/apps/web/package.json
PauloMFJ: I agree! Have updated based on this reflection :)

dataweaver/apps/web/src/app/not-found.tsx
nick-nlb: Right! That's a good point re Next guidelines and routing.

Let's go with option 1 here: dashes for routing, and snake_case for every other file. 

The thinking here is that we may incorporate / re-use / rework features that we build here in other contexts in the future that aren't Next.js specific - including potentially the idea of the theming package. Where possible, I would favor the global style guide rather than the frameworks (with any exceptions such as routing as you pointed out).

dataweaver/FRONTEND.md
```diff
@@ -0,0 +1,261 @@
...
+
+| Topic | This project (Google) | Enforced by |
+|---|---|---|
+| Object shapes | **`interface` over `type`** (props named `ComponentNameProps`) | Biome `useConsistentTypeDefinitions: interface` |
+| Indentation | **2 spaces** (no tabs) | Biome / Stylelint |
+| Translucent colour | **`rgb(var(--color-x) / <alpha>)`** — tokens are CSS variables | convention |
+| Token source | **`@package/tokens`** (`/css`, `/scss`, `/ts`) — not in-app generated | convention |
+| Class names | **kebab-case** | Stylelint `selector-class-pattern` |
+
+Full rule sources: [Google TypeScript Style Guide](https://google.github.io/styleguide/tsguide.html)
```
nick-nlb, line 18: Thank you for adding these files in!

dataweaver/apps/web/src/styles/tokens/_generated.module.scss
nick-nlb: I do like the restricted color palette for consistency. However, because we have a requirement that the application be themable, if we expose those primitive color names directly to the application, the themability starts to break down. 

For example, if we have "--color-blue-light" as an exposed token, and the partner wants that to be purple, you end up having a mismatch between token names and what they represent. An example might be where all the buttons in our app are DC blue, but the partner wants them to be purple.  Or where a partner wants to build a darker theme where the canvas is no longer white.

Because of this, we should expose semantic tokens at the app layer. We can get the advantages of both named color palettes and semantic tokens with a two-tiered token architecture:
* Primitive tokens: the literal, restricted color palette
* Semantic tokens: the contextual application of those primitives (this is what the app sees)

For example, within the package you could define the base palette as you do now:

```
{
   "white": [255, 255, 255]", 
   "blue-light: [247, 252, 255]",
   ... etc,
}
```

Then (still inside the package) we map those to our semantic tokens
```
{
  "primary-main": palette["blue"],
  "primary-light": palette["blue-light"],
  "canvas-background": palette["white"],
  "canvas-decorators": palette["blue-light"],
  ... etc
 }
 ```
 
That way, we still have our restricted consistent color set for developing the base theme, but partners can safely override aspects of the application semantically.

dataweaver/biome.json
nick-nlb: Thank you! This looks good (and if we spot anything else later we can adjust as we go).

nick-nlb (changes requested):
Thank you very much for the quick turnaround Paulo!

I've added in some follow-up comments and responses here.

dataweaver/apps/web/src/styles/tokens/_generated.module.scss
nick-nlb: A follow-up thought on this: 

If we have the package produce component level tokens (i.e. `canvas-background`) directly, then it becomes highly coupled to this particular application. 

If we have it produce specific color primitives, then it becomes a very narrow package for producing specific named colors that require high coupling to those colors on the application side.

As a balanced boundary, we can treat the package's output as a high-level semantic design system. Instead of component-specific tokens, the package can export semantic definitions such as:
* `color-primary-main`
* `color-surface-base` (for generic backdrops like the canvas)
* `color-text-primary`

The application can then import these semantic tokens and either use them directly or map them to internal component variables (either of which could be overridden to make a custom theme).

We then have a clear theme contract. We can still manage the primitive color palette internally within the package, but on the application side, partners can override the semantic tokens to build custom themes without the package needing to know what a "canvas" is, or the application ever seeing "color-blue-light".

Let me know what you think of this.

dataweaver/apps/web/src/app/not-found.tsx
PauloMFJ: Fair - updated!

dataweaver/apps/web/src/styles/tokens/_generated.module.scss
PauloMFJ: Good to know we have the theming requirement. With that in mind I quite like the suggestion of a 'high-level semantic design' system here. That way color names aren't directly tied to the color value and more of a use-case. Should make things quite easy to extend in the future.

Have updated with this in mind! We can also keep an eye on it and iterate as we see fit 💪

nick-nlb (approved):
Thank you for the updates Paulo!

Something to add to the radar (not for this PR):

We should add a step into our CI process to run the token generator (in order to make sure that on commit/PR we never have modified sources that don't match the CSS, TS, etc, outputs).

PauloMFJ:
> Thank you for the updates Paulo!
> 
> Something to add to the radar (not for this PR):
> 
> We should add a step into our CI process to run the token generator (in order to make sure that on commit/PR we never have modified sources that don't match the CSS, TS, etc, outputs).

I agree! In other projects I've also gitignored the token output dist folder alongside a `prepare` script. May be worth doing here too when we tackle that.

beets (approved):

PR #352, opened 2026-06-01
Implement Initial Atlas

PauloMFJ:
## Overview

Implements the initial **Atlas** — an infinite, tldraw-backed canvas that renders data "cards" laid out on a grid. This establishes the rendering foundation and the component architecture for the app, ahead of wiring up a real data source.


dataweaver/apps/web/src/components/elements/card/chart/data_line_chart.tsx
```diff
@@ -0,0 +1,51 @@
...
+const GRID_COLOR = `rgb(${COLORS['surface-decorator']})`;
+const AXIS_COLOR = `rgb(${COLORS['card-content-muted']})`;
+
+interface ChartProps {
+  data: ChartDatum[];
+  width: number;
+  height: number;
+}
+
+export const DataLineChart = ({ data, width, height }: ChartProps) => {
```
nick-nlb: nit: 

I like the principle of category-first naming that you outlined in the FRONTEND.md file. 

DataChartLine? 

(I suppose you could think of "line chart" as a single thing here, but since we do have a category narrowing of "data_table", maybe "data_chart_x" makes sense here too.

dataweaver/apps/web/src/components/scopes/atlas/atlas.tsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export const AtlasProvider = ({ children }: AtlasProviderProps) => {
+  const editorRef = useRef<Editor | null>(null);
+  const pendingShapesQueueRef = useRef<((editor: Editor) => void)[]>([]);
+
+  // TODO: For now using count for positioning - we likely want to use a smarter
+  // approach that accounts for deleted content and doesn't rely on order of
+  // addition, etc later once we hook up to real data
+  const countRef = useRef(-1);
```
nick-nlb: This will change later as per the comment, but what why is this initialized to -1 here?

It looks like there is, at least theoretically, the ability for content to be added before the component is mounted, and so for the the count to be used before it is set to 0 on mounting  (`countRef.current = 0`).  You would get both a -1 as the starting index, but then also potentially items colliding on the same index as the index is reset.

(This might really be a question more of the intention behind the flow here).

nick-nlb (dismissed):
Thanks for this Paulo! 

A couple of comments (one is a nit) but otherwise looks good to go.

beets (dismissed):

dataweaver/apps/web/src/components/elements/card/chart/data_line_chart.tsx
PauloMFJ: I agree with you!

dataweaver/apps/web/src/components/scopes/atlas/atlas.tsx
PauloMFJ: It was -1 as we did countRef.current++ at start of the mount function. I've tweaked so that it can be 0 starting.

But the flow will be reworked anyway once we implement a more robust positioning system. The current flow purely just allows us to add items in a row based on latest index x fixed card width.

nick-nlb (approved):

nick-nlb (approved):

PR #355, opened 2026-06-03
Implement Atlas Controls

PauloMFJ:
## Implement Atlas Controls

Adds the floating control chrome for the Atlas (tldraw) canvas — a zoom control, export button, and a tool palette — plus the supporting design-system pieces (icons, tokens, typography, a `Button` size variant).

### Controls component
- New `scopes/atlas/components/controls.tsx` + `.module.scss`, rendered through tldraw's `InFrontOfTheCanvas` slot so it reads live editor state via `useEditor` while staying out of the canvas transform.
- **Zoom**: in/out buttons + a live zoom readout. The displayed value is rescaled from the real zoom range onto a friendlier `0–200` scale via a new `mapRange` helper, and steps by a constant amount.
- **Tool palette**: Select / Pan / Text, bound 1:1 to tldraw tool ids (`select` / `hand` / `text`), with the active tool highlighted.
- **Export**: button in place (wiring to the export flow is a follow-up — see TODO below).
- All three control groups are built on the shared `Button` element with per-state `colorScheme`s.

### Atlas editor config
- New `scopes/atlas/config.ts` centralizes the zoom config: `MIN_ZOOM` / `MAX_ZOOM` and an evenly-spaced `ZOOM_STEPS` array generated from a single display-step value (keeps the on-screen increment constant), plus the display range.
- `atlas.tsx` now caps the camera zoom (`setCameraOptions`), styles the built-in text tool to match the design system, and tints the selection stroke from the accent token (`updateThemes`). Selection colours are also wired through `--tl-color-*` overrides in `atlas.module.scss`.

### Design system
- **Icons**: `cursor`, `hand`, `insert_text`, `minus`, `plus`.
- **`Button`**: new required `size` prop (`'small' | 'large'`) → `data-size`.
- **Tokens** (`colors.json`): added `control-surface`,
  `control-surface-hover`, `control-accent`, `control-accent-content`,
  `control-content`, and `surface-content`; renamed `card-shadow` → `shadow`.
- **Typography**: renamed the mixins onto a clearer scale —
  `type-label → type-label-large`, `type-tab → type-label-small` and unified previous `type-tab` with `type-label-small`; all usages
  updated.

### Follow-ups
- `Controls` export button is a no-op (`// TODO: Support export here`) — needs wiring to the canvas export API.

### Testing
- `pnpm lint` (type-check + Biome + Stylelint) and `pnpm build` pass.
- Verified controls render over the canvas; zoom in/out, tool switching, and active-tool highlighting work; selection stroke uses the accent colour.


dataweaver/apps/web/src/components/scopes/atlas/config.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * The discrete zoom levels the controls step through, generated so they stay
+ * evenly spaced across [MIN_ZOOM, MAX_ZOOM] — that even spacing is what makes
+ * the displayed value increment by a constant `ZOOM_DISPLAY_STEP`.
+ */
+const ZOOM_STEP_COUNT =
+  (ZOOM_DISPLAY_RANGE[1] - ZOOM_DISPLAY_RANGE[0]) / ZOOM_DISPLAY_STEP;
+
+/**
+ * The discrete zoom levels the controls step through, generated so they stay
```
nick-nlb, line 37: These two JS docs are identical. The one for `ZOOM_STEP_COUNT` could be updated to say something like:

`The number of discrete zoom steps between MIN_ZOOM and MAX_ZOOM.`

dataweaver/apps/web/src/styles/includes/_typography.module.scss
```diff
@@ -12,16 +12,18 @@
...
-  font-size: 12px;
+  font-size: 20px;
   font-weight: 500;
-  line-height: 1.33;
+  line-height: 1;
+  text-box: trim-both cap alphabetic;
 }
 
-@mixin type-tab {
+@mixin type-label-small {
```
nick-nlb, line 23: The description mentions that `type-label-medium` is a newly relabeled type (but doesn't appear here). Similarly, the description references type-control, but I don't think that was there.

This might be just be a case of updating the description to bring it in sync with the changes as they are now.

dataweaver/apps/web/src/components/scopes/atlas/config.ts
```diff
@@ -0,0 +1,48 @@
...
+ * evenly spaced across [MIN_ZOOM, MAX_ZOOM] — that even spacing is what makes
+ * the displayed value increment by a constant `ZOOM_DISPLAY_STEP`.
+ */
+export const ZOOM_STEPS: readonly number[] = Array.from(
+  { length: ZOOM_STEP_COUNT + 1 },
```
nick-nlb, line 42: We could do some thinking on what the zoom levels are, and how actual zoom levels map onto the displayed zoom levels (i.e. the percentages) in the toolbar.

Right now, we start at 50% (at a level that might be more intuitive as 100%) and the actual neutral zoom factor of 1 (between 0.25 and 3.25) is skipped by the zoom steps. You also can't get back to the starting 50%.

However, this could be find-tuned after this PR (but want to flag so we don't lose sight of it).

nick-nlb (dismissed):
Thank you Paulo! Some comments, but they don't have to hold up the merging of this (however, the ones we don't do here, coud we flag to be addressed for a future PR)?

dataweaver/apps/web/src/components/scopes/atlas/config.ts
PauloMFJ: I agree! If you have any thoughts on exactly how you / the team would like the scaling to work I can then address in its own PR 👌

dataweaver/apps/web/src/styles/includes/_typography.module.scss
PauloMFJ: Ah yes - good spot updated!

nick-nlb (approved):

PR #357, opened 2026-06-08
Implement Error Toasts

PauloMFJ:
## Overview

Adds a lightweight in-house toaster (foundations/toaster) based on [Sonner](https://www.npmjs.com/package/sonner).

- Stacked toasts that collapse into a scaled pile and fan out on hover/focus, with enter/exit animations and pause-on-interaction
- Click anywhere to dismiss (keyboard-accessible) + auto-dismiss timer
- a11y: labelled aria-live region, per-toast dismiss labels, focus-aware expand/pause
- Reduced motion: identical behaviour, just without durations
- Adds toaster-* color tokens, a `type-headline` mixin, and a `$z-index-toaster` token
- Fixes the card export / view-chart actions (onclick → onClick) that silently never fired

## Screenshots

image
image


dataweaver/apps/web/src/components/elements/button.tsx
```diff
@@ -33,6 +35,7 @@ type ButtonProps = {
 export const Button = ({
   icon: Icon,
   children,
+  size,
```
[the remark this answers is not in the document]
PauloMFJ: I disagree. I went back and edited all previous instances; leaving as is.

dataweaver/apps/web/src/components/foundations/toaster/toaster.tsx
```diff
@@ -0,0 +1,209 @@
...
+    // Newest first so index 0 is the front of the stack.
+    const orderedToasts = [...toasts].reverse();
+
+    // Running total of the heights of the toasts in front of the current one;
+    // ends the loop holding the height of every toast combined
+    let heightsBefore = 0;
+    const populatedToasts = orderedToasts.map((toast, index) => {
+      const y = -(ROW_GAP * index) - (isExpanded ? heightsBefore : 0);
+      const scale = isExpanded ? 1 : 1 - SCALE_STEP * index;
+      const opacity = isExpanded || index < MAX_VISIBLE_TOASTS ? 1 : 0;
```
nick-nlb, line 161: We are setting the opacity here to 0 when the toasts are hidden, but they are still rendered on the canvas (above and staggered behind the visible toasts).

This means that hidden toasts will still interfere with click events on cards/the canvas behind them (or in practice, mousing over the hidden items will trigger the stack to go into the non-collapsed view. 

Assuming that wasn't the intended behavior, could we add a `pointer-events` of `none` here to make the hidden ones completely inert?

dataweaver/apps/web/src/components/foundations/toaster/store.ts
```diff
@@ -0,0 +1,63 @@
...
+    this.#emit();
+  };
+
+  #emit = () => {
+    for (const listener of this.#listeners) listener();
+  };
+}
+
+/** Singleton store instance used to manage toasts across the app. */
+export const TOAST_STORE = new ToastStore();
```
nick-nlb, line 53: While this currently looks safe as it is because it only ever gets used on the client side, it feels fragile to rely on module state. If toast is ever called accidentally during a server-side render, we could get cross-request contamination.  

There's no case of that now, but thinking forward both in terms of this app, but also in terms of reuse of the toast library later (where it might be more of an issue).

What do you think of moving the toast functionality to a React context and completely decoupling it from the server? (This would be a fairly big refactor and so could be a separate PR). In the meantime we could put a guard that prevents toasts from being added if window isn't present.

nick-nlb (commented):
We can discuss the bigger comment and keep it for another PR, but otherwise the one smaller behavioral comment.

Do you mind also taking some screenshots and including them in the description here, and then pinging miss-o-soup to have a look (in case this is completely new UI/UX). Alternatively, since we're meeting tomorrow anyway, you could give it a demo then.

dataweaver/apps/web/src/components/foundations/toaster/store.ts
PauloMFJ: I considered using a React context but actually ended up choosing this path to match closer to Sonner's implementation + general DX improvements over needing hook. Happy to put in a guard to prevent toasts being added if window isn't present in the meantime.

Happy to review / move towards context later if it becomes an issue?

dataweaver/apps/web/src/components/foundations/toaster/toaster.tsx
PauloMFJ: Good spot!

dataweaver/apps/web/src/components/foundations/toaster/store.ts
nick-nlb: That sound good (guard now, and flag it for consideration later).

nick-nlb (approved):
Thanks Paulo!

PR #358, opened 2026-06-08
Implement Initial Prompt w/ Intro

PauloMFJ:
## Overview

Adds the home-screen entry point for the Atlas: a prompt bar and an initial-mount welcome overlay above it.

- Prompt (scopes/page_home/prompt.tsx) — auto-growing textarea (field-sizing, capped at 4 rows) with a visually-hidden label for a11y, Enter-to-submit (Shift+Enter for newline), and a send Button.
- Intro overlay (scopes/page_home/intro.tsx) — dismissible first-run card laid out with CSS grid (close button in its own column), an intro blurb, clickable example prompts, and a "progress isn't saved" notice. Picking an example adds the query to the prompt - it can then be manually submitted.
- page_home promoted to a folder and now composes the prompt + intro, owning the query/intro state.
- Supporting: new arrow_up / close / info icons; _reset.scss now neutralizes textarea; new typography mixins (type-display, type-body-large, type-input); --shadow-elevated + button-size CSS vars in _root.scss; added card-base-subtle token.

## Screenshots

image


dataweaver/apps/web/src/components/primitives/screen_reader.tsx
```diff
@@ -1,18 +1,23 @@
...
   'className'
 >;
 
-export const ScreenReaderOnly = ({
+export const ScreenReaderOnly = <TElement extends ElementType = 'span'>({
+  element,
   children,
   ...rest
-}: ScreenReaderOnlyProps) => {
+}: ScreenReaderOnlyProps<TElement>) => {
```
nick-nlb, line 15: This extends `ComponentPropsWithRef` but doesn't take the ref and pass it through to the Container (not a breaking change now but could be if a consumer passes in a ref that is then dropped).

dataweaver/apps/web/src/components/primitives/icons/arrow_up.tsx
```diff
@@ -0,0 +1,18 @@
...
+  return (
+    <svg
+      {...props}
+      viewBox="0 0 24 24"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      aria-hidden="true"
+    >
+      <path
+        fill="currentcolor"
```
nick-nlb: currentcolor -> currentColor

(This won't cause any functional issues since it is functionally case insensitive but in React camelCase)

dataweaver/apps/web/src/components/scopes/page_home/intro.tsx
```diff
@@ -0,0 +1,78 @@
...
+            'base-hover': 'var(--color-card-base-subtle)',
+            content: 'var(--color-card-content)',
+            'content-hover': 'var(--color-card-content)',
+          }}
+          onClick={onClose}
+        />
+
+        <div className={s['content-container']}>
+          <h2 className={s.title}>
+            Lets discover what you can do with Data Commons
```
nick-nlb: nit: Lets -> Let's

An aside, but we should put i18n on the cards for a future discussion.

dataweaver/apps/web/src/styles/includes/_typography.module.scss
```diff
@@ -1,29 +1,78 @@
+@use "sass:list";
+
+@mixin -font-weight-sans($font-weight, $supportedFontWeights) {
+  @if not list.index($supportedFontWeights, $font-weight) {
+    @error "Invalid font name used '#{$font-weight}'. Please use one of the supported weights - '#{$supportedFontWeights}'.";
+  }
+
+  // Note: If adding any new font weights in here, make sure to also add
+  // respective weights to the 'FONT_URL' in 'layout.txt' file
```
nick-nlb: Very minor nit: layout.tsx

dataweaver/apps/web/src/components/scopes/page_home/intro.tsx
```diff
@@ -0,0 +1,78 @@
...
+      transition={{ duration: 0.1, ease: EASE_LINEAR }}
+    >
+      <div className={s['inner-container']}>
+        <Button
+          className={s['button-close']}
+          icon={IconClose}
+          size="medium"
+          aria-label="Close"
+          colorScheme={{
+            base: 'transparent',
```
nick-nlb, line 36: Since the color values are rgb(a), this would turn into rgb('transparent'). `0 0 0 / 0%` or similar instead?

dataweaver/apps/web/src/components/scopes/page_home/page_home.tsx
```diff
@@ -1,13 +1,32 @@
...
   useEffect(() => {
+    return () => {
+      // Clear any in-flight timeouts if the page unmounts mid-response
+      for (const timeout of timeoutsRef.current) clearTimeout(timeout);
+    };
+  }, []);
+
+  // TODO: This is entirely using mock data for now
+  const runMockQuery = () => {
+    for (const timeout of timeoutsRef.current) clearTimeout(timeout);
```
nick-nlb, line 27: This is a mock, and not an issue but:

If you submit a new query while a current query is pending (i.e, a mock is waiting for the current timer) the timer is cleared and the loading cards get stuck in limbo. 

We should start to think about what the functionality should be here when this starts submitting real requests (should we not be able to submit a new request until the current one resolves, should subsequent requests abort current ones, etc).

nick-nlb (approved):
Thanks Paulo. It's looking great.

Some comments (mostly nits), to address before submission but otherwise approved.

nick-nlb:
A follow-up comment:

The expanding text field works well. For you and Adriana to consider: the dotted line around it when it is in focus (particularly in the context of the canvas), gives me the instictual feeling that I can drag it around. Maybe because it looks like the Photoshop text boxes.  

I do like that there is an indication of you being in the text box, but we could brainstorm options here.

nick-nlb:
A screenshot for context:

image

dataweaver/apps/web/src/components/primitives/screen_reader.tsx
PauloMFJ: It does right? If consumer uses `ref` then it'll work as expected. In past with `forwardRef` this wouldn't have been the case but `rest` here will expose `ref` unless I'm missing something?

dataweaver/apps/web/src/components/scopes/page_home/intro.tsx
PauloMFJ: Good spot - I suppose it just breaks which is why it works as intended right now. Will fix

dataweaver/apps/web/src/components/scopes/page_home/page_home.tsx
PauloMFJ: Agreed. TBF for now only added this to avoid spamming timeouts - but in the near future once we have real data we'll defo have to cross that line of how cards resolve in parallel.

dataweaver/apps/web/src/styles/includes/_typography.module.scss
PauloMFJ: Eagle eye!

PauloMFJ:
> A follow-up comment:
> 
> The expanding text field works well. For you and Adriana to consider: the dotted line around it when it is in focus (particularly in the context of the canvas), gives me the instictual feeling that I can drag it around. Maybe because it looks like the Photoshop text boxes.
> 
> I do like that there is an indication of you being in the text box, but we could brainstorm options here.

Good point. Will raise with Matt too to see if he has any design ideas for a better focus state across the app.

PR #362, opened 2026-06-10
Implement Initial Export Panel

PauloMFJ:
## Overview

Adds a floating Export panel to the Atlas canvas, triggered from the existing Export control.

### What changed
- Export panel — new dropdown (`export_panel.tsx`) rendered in the `InFrontOfTheCanvas` slot: header, a status card, and an "Export options" grid (API Request Code, CSV, AI Narrative, AI Infographic, SVG, PNG). Opens with a prefers-motion-gated fade/scale and stops pointer/wheel events so canvas gestures don't fire underneath.
- Selection-aware empty states — derives its state live from the canvas: no cards ("Create a prompt to begin"), cards but none selected ("No cards selected" + a Select All shortcut), or ready ("N cards selected").
- Export context — `ExportProvider` / `useExport` (open/close/toggle) colocated in one file, wrapping the canvas and page so the panel can be opened from anywhere; the Export button in `controls.tsx` now toggles it with aria-haspopup/aria-expanded.
- New icon primitives — close, select, shapes.

## Screenshots

image


dataweaver/apps/web/src/components/primitives/icons/close.tsx
```diff
@@ -0,0 +1,18 @@
...
+  return (
+    <svg
+      {...props}
+      viewBox="0 0 24 24"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      aria-hidden="true"
+    >
+      <path
+        fill="currentcolor"
```
nick-nlb: Nit: Same comment as the other branch: currentcolor -> currentColor

dataweaver/apps/web/src/components/scopes/atlas/components/in_front_of_canvas/export/export_panel.module.scss
```diff
@@ -0,0 +1,132 @@
+.container {
+  position: absolute;
+  top: 80px;
+  right: 14px;
+  display: flex;
+  flex-direction: column;
+  width: 360px;
+  padding: 20px;
```
nick-nlb, line 14: It might not be a trivial change now, but we should put into the task list the conversion of these hardcoded values to theme level spacing constants.

dataweaver/apps/web/src/configs/environment_client.ts
```diff
@@ -1 +1,5 @@
 export const IS_BROWSER = typeof window !== 'undefined';
+
+export const IS_IOS =
```
nick-nlb: Small nit: the constant checks for Mac in addition to the i * family. Maybe IS_MAC or IS_APPLE?

dataweaver/apps/web/src/styles/includes/_typography.module.scss
```diff
@@ -1,17 +1,49 @@
-@mixin type-title {
+@use "sass:list";
+
+@mixin -font-weight-sans($font-weight, $supportedFontWeights) {
+  @if not list.index($supportedFontWeights, $font-weight) {
+    @error "Invalid font name used '#{$font-weight}'. Please use one of the supported weights - '#{$supportedFontWeights}'.";
```
nick-nlb: Nit: invalid font weight (rather than name)

dataweaver/apps/web/src/components/scopes/atlas/components/in_front_of_canvas/export/export_panel.tsx
```diff
@@ -0,0 +1,195 @@
...
+      .map((shape) => shape.id);
+    editor.select(...cardIds);
+  };
+
+  return (
+    <AnimatePresence>
+      {isOpen && (
+        <m.section
+          className={s.container}
+          role="dialog"
```
nick-nlb, line 95: When this card opens, in keeping with dialog-like behavior of it, can we bring focus into the dialog? (also, more generally, we will need to implement keyboard navigation for this). This can be a separate PR, and might be part of a more general discussion about how that navigation should work.

nick-nlb (approved):
Thanks for this as well Paulo! I've put in a couple of nits, but otherwise approved.

One of the items out of the last meeting was that the "status" section of the export (where it says "No cards selected") is functionally useful but should be smaller. We can put that in for a separate PR (along with the other deferred item mentioned in the comments. 

We should make sure to track the various differed items that come out of this and other PRs so that we don't lose sight of them.

dataweaver/apps/web/src/components/scopes/atlas/components/in_front_of_canvas/export/export_panel.tsx
PauloMFJ: Defo. Have added focus trap hook + escape key support and also good point. I've been meaning to review how tabbing works in relation to TLDraw. Have noticed the latter seems to consumes a lot of tab events and makes it quite hard to break point into surrounding UI.

dataweaver/apps/web/src/components/scopes/atlas/components/in_front_of_canvas/export/export_panel.module.scss
PauloMFJ: I agree - added to my list!

PR #364, opened 2026-06-15
Reduce Intro Sizings

PauloMFJ:
Before:
image

Intermediate (out of date):
image

After:
304-pre-post

### DIFF
--- a/dataweaver/apps/web/src/styles/includes/_typography.module.scss
+++ b/dataweaver/apps/web/src/styles/includes/_typography.module.scss
@@ -40,7 +40,7 @@ $-supported-title-font-weights: 500, 700;
 
   font-family: "Google Sans", sans-serif;
   font-size: 16px;
-  line-height: 1.2;
+  line-height: 1.4;
 }
 
 $-supported-body-font-weights: 400, 500;
@@ -53,7 +53,7 @@ $-supported-body-font-weights: 400, 500;
 
   font-family: "Google Sans", sans-serif;
   font-size: 18px;
-  line-height: 1.2;
+  line-height: 1.4;
 }
 
 @mixin type-body-small($font-weight: 400) {
@@ -64,14 +64,14 @@ $-supported-body-font-weights: 400, 500;
 
   font-family: "Google Sans", sans-serif;
   font-size: 14px;
-  line-height: 1.2;
+  line-height: 1.4;
 }
 
 @mixin type-label-large($has-text-box-trim: true) {
   font-family: "Google Sans", sans-serif;
   font-size: 20px;
   font-weight: 500;
-  line-height: 1.2;
+  line-height: 1.4;
 
   @if $has-text-box-trim {
     text-box: trim-both cap alphabetic;
@@ -82,7 +82,7 @@ $-supported-body-font-weights: 400, 500;
   font-family: "Google Sans", sans-serif;
   font-size: 16px;
   font-weight: 500;
-  line-height: 1.2;
+  line-height: 1.4;
 
   @if $has-text-box-trim {
     text-box: trim-both cap alphabetic;
@@ -93,7 +93,7 @@ $-supported-body-font-weights: 400, 500;
   font-family: "Google Sans", sans-serif;
   font-size: 12px;
   font-weight: 500;
-  line-height: 1.2;
+  line-height: 1.4;
 
   @if $has-text-box-trim {
     text-box: trim-both cap alphabetic;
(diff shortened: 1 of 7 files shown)


dataweaver/apps/web/src/styles/includes/_typography.module.scss
```diff
@@ -64,14 +64,25 @@ $-supported-body-font-weights: 400, 500;
...
-  line-height: 1.5;
+  line-height: 1.2;
 }
 
 @mixin type-label-large($has-text-box-trim: true) {
   font-family: "Google Sans", sans-serif;
   font-size: 20px;
   font-weight: 500;
-  line-height: 1.4;
+  line-height: 1.2;
```
nick-nlb: This change has an effect on the prompt input (because of the line `max-height: calc(var(--max-rows) * 1lh);`). With the lowering of the line-height, on my browser (but not in the screenshot), there is now a scroll bar. 

If you can't see t on your side and you are using a Mac and a touch scrolling device, it might be suppressing scrollbars (System Settings -> Appearance -> Show scroll bars, set to "Always" temporarily to see the issue).

nick-nlb: I had actually started to review this before your latest batch of changes, and so the above comment is now out-of-date (in that the scroll doesn't appear). I'll still keep it for your interest, as (if I am correct that you are using a touch scrolling device) then it's an interesting problem that has affected me before, where you don't see what other people see because of how the Mac is choosing to render or not render scrollbars.

nick-nlb:
The after screenshot looks like an intermediary screenshot before the final round of changes. Could we update so @miss-o-soup can see the final version?

dataweaver/apps/web/src/styles/includes/_typography.module.scss
```diff
@@ -52,8 +52,8 @@ $-supported-body-font-weights: 400, 500;
   );
 
   font-family: "Google Sans", sans-serif;
-  font-size: 20px;
-  line-height: 1.5;
+  font-size: 18px;
```
nick-nlb, line 55: It's not part of the scope of this PR, but can we add to the list an item to "de-pixel" the font sizes (converting them to rem/ems) to support user-customized base sizes for accessibility. For our own DX, if we still want to think of them as pixels, we could add conversion functions in the scss that converts from a pixel value as we see it to a rem, based on a default of 16px.

nick-nlb (approved):
This looks good (pending the comment to update the screenshots for  @miss-o-soup, as well as adding the pixel to rem task into the to-do list).

nick-nlb:
> The after screenshot looks like an intermediary screenshot before the final round of changes. Could we update so @miss-o-soup can see the final version?

@miss-o-soup  I've updated the screenshot with a before and after GIF (because its easier to see the difference between the two). The bigger version is the original version, and the smaller version is what it looks like after this PR.

nick-nlb:
> > The after screenshot looks like an intermediary screenshot before the final round of changes. Could we update so @miss-o-soup can see the final version?
> 
> @miss-o-soup I've updated the screenshot with a before and after GIF (because its easier to see the difference between the two). The bigger version is the original version, and the smaller version is what it looks like after this PR.

@PauloMFJ  One thing you can see from the animation between the two is that the closing "X" no longer lines up with the header. It's not a given that they should, and this may be intentional and line-up with design specs, but flagging in case.

miss-o-soup:
Thank you for adding the screenshots and the updates here.
I see how the intro screen got updated and I love the detail on the line height in the "After" screen.
Thank you! - looks good for me! :-)

dataweaver/apps/web/src/styles/includes/_typography.module.scss
PauloMFJ: Good point! Will keep an eye out / test with scrollbars enabled when I do further work on the prompt :)

dataweaver/apps/web/src/styles/includes/_typography.module.scss
PauloMFJ: Cool - added it to my list :)

PauloMFJ:
> > > The after screenshot looks like an intermediary screenshot before the final round of changes. Could we update so @miss-o-soup can see the final version?
> > 
> > 
> > @miss-o-soup I've updated the screenshot with a before and after GIF (because its easier to see the difference between the two). The bigger version is the original version, and the smaller version is what it looks like after this PR.
> 
> @PauloMFJ One thing you can see from the animation between the two is that the closing "X" no longer lines up with the header. It's not a given that they should, and this may be intentional and line-up with design specs, but flagging in case.

Good spot! And correct - noticed it was actually un-aligned in the designs so updated to match; now the 'X' is equally distanced on top / right edge. The content then positioned with enough space to give it room to brief above the title.

PR #369, opened 2026-06-17
Implement Context Menu Overrides w/ Font Family Fix

PauloMFJ:
## Before
Right click without selected card(s):
image

Right click with selected card:
image

Right click with selected cards:
image

## After
Right click without selected card(s):
image

Right click with selected card:
image

Right click with selected cards:
image


dataweaver/apps/web/src/components/scopes/atlas/atlas_provider.module.scss
```diff
@@ -5,6 +5,14 @@
   --tl-color-selected: rgb(var(--color-control-accent));
   --tl-color-selected-contrast: rgb(var(--color-control-accent-content));
 
+  // The UI font is inherited from 'html' (see styles/core/_base.scss). These
+  // align tldraw's own font CSS vars brand font so they don't fall back
+  // to the bundled tldraw_* fonts
+  --tl-font-draw: "Google Sans", sans-serif;
+  --tl-font-sans: "Google Sans", sans-serif;
+  --tl-font-serif: "Google Sans", serif;
+  --tl-font-mono: "Google Sans", monospace;
```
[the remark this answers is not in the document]
PauloMFJ: To keep to Brand IMO it's better to just override - we don't plan on ever natively supporting these features so it's not a factor we should factor into at this stage.

dataweaver/apps/web/src/styles/core/_base.scss
```diff
@@ -1,6 +1,11 @@
 @use "~/styles/includes";
 
 @layer base {
+  html {
+    // Set global default font-family
+    font-family: "Google Sans", sans-serif;
```
nick-nlb, line 6: The font will need to be tokenized and swappable like other elements. 

We can note and track the fact that we will need to be able to provide an entry point for partners who wish to customize by overriding the tokens and provide their own imports (we can discuss soon). But then regardless, the font should be a token along with other theme variables. 

We don't have to apply that in this PR (we could save that for a bigger look at how theming is applied in practice), but if not, can we note it in your tracker?

nick-nlb (approved):

dataweaver/apps/web/src/styles/core/_base.scss
PauloMFJ: Will note this down!

PR #373, opened 2026-06-22
Overhaul Color Tokens Setup w/ Extended Button

PauloMFJ:
## Overview

Reworks the design-token system into a clear two-layer model and rebuilds the Button's color API on top of it.

### Tokens (packages/tokens)
- Two layers: generic roles (accent, surface, content, shadow, …) hold the real values; component-scoped sets (button-*, card-*, control-*, atlas-*, query-*, radio-*, …) alias them via "$role". Components reference only their own scope's tokens — nothing borrows generics or sibling scopes directly (one deliberate exception: the button control / card-action tones).
- Value forms: RGB triplet [r,g,b], triplet + alpha [r,g,b,a] (emitted as r g b / a), or "$alias".
- Generator fixes: COLORS (TS) export now carries alpha in the same r g b / a form as the CSS vars; removed a channel-array mutation bug; alpha 0 no longer rejected.
- One-off values are inlined in their component token rather than added to the global palette; pruned dead tokens.

### Button
- New API: variant (flat | border) + tone (subtle, subtle-highlight, prominent, accent, accent-subtle, control, card-action).
- Each tone is a self-contained token set (base / content / border + hovers) — retune one tone without affecting others.
- Explicit per-tone border / border-hover channels (the 5 standalone tones; control / card-action excluded).
- Disabled state is derived from the active tone (dropped the separate disabled-style prop).

### Docs
- AGENTS.md updated to document the role/scope layering, value forms (incl. alpha + $alias), and the one-off-inline rule.

### Build
- Generated token output (`packages/tokens/dist/`) is no longer committed — added
  `packages/tokens/.gitignore` (`dist/`) and a root `prepare` script
  (`pnpm run generate:tokens`) so it's regenerated on install / build.

### New Buttons

Prompt and Status buttons before / after:

image
image

Export button before / after:

image
image

Everything else matches existing existing styles visually.


dataweaver/apps/web/src/components/scopes/atlas/components/in_front_of_canvas/export/control.tsx
```diff
@@ -21,7 +14,8 @@ export const Control = ({ id, isOpen, onToggle }: ControlProps) => {
       className={s['button-export']}
       icon={IconExport}
       size="large"
-      colorScheme={BUTTON_EXPORT_COLOR_SCHEME}
+      variant="flat"
+      tone="control"
       aria-haspopup="dialog"
       aria-expanded={isOpen}
       aria-controls={id}
```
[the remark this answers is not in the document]
PauloMFJ: It's used with children.

dataweaver/packages/tokens/src/generate.ts
```diff
@@ -55,9 +104,9 @@ const generateTypeScript = (): string => {
   // Colors
   lines.push('');
   lines.push('export const COLORS = {');
-  for (const [name, values] of Object.entries(COLORS)) {
+  for (const name of Object.keys(COLOR_MAP)) {
     const constName = name.includes('-') ? `'${name}'` : name;
-    lines.push(`  ${constName}: '${values.join(', ')}',`);
+    lines.push(`  ${constName}: '${toCssChannels(resolveChannels(name))}',`);
```
nick-nlb: Differently to the CSS version, this one resolves to the actual color rather than the name. So where the CSS file has:
` ``
 --color-atlas-surface: var(--color-surface);
```
The TS file has 
 ```
 'atlas-surface': '247 252 255',
```

This means that if the partner overrides --color-surface, it will affect only CSS/SCSS but not the TS versions. Let's make this resolve to the variable in the same way.

nick-nlb (approved):
Thanks Paolo! The refactoring of the theme looks great (as does the reworking of the button variants).. I like the separation of token layers the overall refactor introduces. 

One comment to address but approved.

dataweaver/packages/tokens/src/generate.ts
PauloMFJ: Updated TS to resolve with shared consts for any inline values - let me know if this wasn't what you were intending?

PR #374, opened 2026-06-22
Implement Prompt Tags Support

PauloMFJ:
## Overview

Implement prompt tags support.

With 1 tag:
image

With 2 tags:
image

With 3+ tags:
image


dataweaver/apps/web/src/components/elements/tag.module.scss
```diff
@@ -0,0 +1,18 @@
+.container {
+  display: flex;
+  column-gap: 8px;
+  align-items: center;
+  height: var(--tag-height);
+
+  // Note: The extra 1px here accounts for the 1px border
+  padding-inline: 12px
+    calc((var(--tag-height) - var(--button-circle-size-extra-small)) / 2 - 1px);
```
nick-nlb: This seems a bit fragile (relying on the size of the border). I played with this line to try achieve the same effect without the complex calculation :`padding-inline: 12px 4px;`.

This gave the same result in the examples that I tried. Of course, if there is another aspect or scenario of this that requires the calc, we can keep it in. Let me know what you think.

dataweaver/apps/web/src/components/scopes/page_home/prompt.module.scss
```diff
@@ -11,6 +12,14 @@
   box-shadow: var(--shadow-elevated);
 }
 
+.tags-inner-container {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 12px;
+  align-items: center;
+  padding-bottom: var(--row-gap);
```
nick-nlb: This is converted to a token, but it's also moved responsibility for the internal spacing of the container's children from the container itself to the children. This creates a less flexible coupling between the children, where the children have to understand the layout of the parent and independently implement it. If there isn't another difference that requires this separation, let's keep the gap handling in the parent.

nick-nlb (approved):
Thank you! 

I put in a couple of comments to look at before merging, but otherwise approved.

One overall aspect of that (more on the design-side) that we can discuss:

The spacing around the tag section seems large, and this makes the expansion / contraction of the box quite noticeable. 

We can discuss with the full designers, and I will of course differ to you all! However, if we made those gutters much thinner (and maybe make distance from top of the card to the text field a bit larger when there are no tags), we might be able to minimize or nullify the need for expanding the card itself when we add tags.

Image

dataweaver/apps/web/src/components/elements/tag.module.scss
PauloMFJ: I agree - I simplified it

dataweaver/apps/web/src/components/scopes/page_home/prompt.module.scss
PauloMFJ: This has been simplified with latest UI. No longer need this nested CSS var

PauloMFJ:
> Thank you!
> 
> I put in a couple of comments to look at before merging, but otherwise approved.
> 
> One overall aspect of that (more on the design-side) that we can discuss:
> 
> The spacing around the tag section seems large, and this makes the expansion / contraction of the box quite noticeable.
> 
> We can discuss with the full designers, and I will of course differ to you all! However, if we made those gutters much thinner (and maybe make distance from top of the card to the text field a bit larger when there are no tags), we might be able to minimize or nullify the need for expanding the card itself when we add tags.
> 
> Image

I agree with this sentiment. I had a conversation with the designer and have updated based on the chats. Let me know if you have any thoughts with latest version. Should feel less jumpy as user selects / de-selects tags now that we are utilising the shared space offered by submit button. Note: I've tweaked some of the prompt spacings with this PR but I still have to do a full pass at reduced sizings across UI. cc @miss-o-soup also let us know what you think!

nick-nlb:
> > Thank you!
> > I put in a couple of comments to look at before merging, but otherwise approved.
> > One overall aspect of that (more on the design-side) that we can discuss:
> > The spacing around the tag section seems large, and this makes the expansion / contraction of the box quite noticeable.
> > We can discuss with the full designers, and I will of course differ to you all! However, if we made those gutters much thinner (and maybe make distance from top of the card to the text field a bit larger when there are no tags), we might be able to minimize or nullify the need for expanding the card itself when we add tags.
> > Image
> 
> I agree with this sentiment. I had a conversation with the designer and have updated based on the chats. Let me know if you have any thoughts with latest version. Should feel less jumpy as user selects / de-selects tags now that we are utilising the shared space offered by submit button. Note: I've tweaked some of the prompt spacings with this PR but I still have to do a full pass at reduced sizings across UI. cc @miss-o-soup also let us know what you think!

image

Here is a screenshot of how it looks now. We can demo it and discuss tomorrow. It definitely feels better without the box needing to expand to accommodate the chips.

PauloMFJ:
Merging in to have it with latest changes ahead of our call. I can address any further discussion points in a separate PR.

PR #375, opened 2026-06-22
Scott/store

sagrimson:
### Summary

Introduces a centralized Zustand store (`AtlasStore`) as the single source of truth for atlas state, replacing the previously scattered local state. Connects the streaming query pipeline end-to-end so queries produce visible cards on the canvas with chart, table, and notes variants.

#### Store (`src/store/`)
- **New `AtlasStore`** (Zustand + devtools) managing history nodes, a card registry, and transient UI state (processing flag, status message).
- **Actions**: `queryStart`, `nodeAddResult`, `queryComplete`, `queryFail`, `queryCancel`, `cardRegister`/`cardUnregister`, etc.
- **Import/export serialization** — download or upload the full store state as a versioned JSON envelope for debugging and sharing.

#### Store ↔ Canvas Sync (`scopes/atlas/sync_store.ts`)
- Derives `AtlasContent` (chart / table / text) from raw `QueryResult` data in the store.
- Syncs store card entries to canvas shapes, adding/removing/updating them reactively.

#### Query Pipeline
- **`QueryProvider`** refactored to dispatch streaming events (`STREAM_EVENT`) into the store rather than managing cards locally.
- **render_result_html.ts** (server) — new step that builds Markdown table + notes HTML from `QueryResult`, parsed via `marked`, so the client receives pre-rendered HTML.
- **Streaming query hook** updated to surface status constants (`STATUS.*`) through the store.

#### UI
- **Facet selector** — dropdown on chart cards to switch between available facets (source/date-range/unit).
- **Table variant** — cards can render as an HTML table alongside chart and text variants.
- **Status component** — shows current prompt + processing status with a cancel button.
- **`HtmlParsed` primitive** — renders trusted server-generated HTML with clickable internal links (triggers new chart queries).
- **Dev panel** — import/export state controls behind dev mode.
- **Home page** — removed duplicate status display, intro no longer hides immediately.

#### Misc
- Markdown parsing moved server-side (avoids shipping `marked` to the client bundle).
- Cards render per-variable-per-place (3 cards per query per place: chart, table, notes).
- `validate_href` utility for safe link handling within parsed HTML.
- `use_dev_mode` hook for toggling developer tooling.


dataweaver/apps/web/src/components/primitives/html_parsed.tsx
```diff
@@ -0,0 +1,66 @@
...
+    () =>
+      html
+        ? parse(html, {
+            replace: (domNode) => {
+              if (domNode instanceof Element) {
+                switch (domNode.name) {
+                  case 'script':
+                  case 'style':
+                  case 'iframe':
+                    return null;
```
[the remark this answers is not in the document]
sagrimson: done

dataweaver/apps/web/src/components/scopes/atlas/sync_store.ts
```diff
@@ -0,0 +1,202 @@
...
+ */
+export const useStoreShapeSync = (atlas: AtlasContextProps) => {
+  const handlesRef = useRef<Map<string, CardHandle<CardVariant>>>(new Map());
+  const prevCardsRef = useRef<Record<string, CardEntry>>({});
+
+  useEffect(() => {
+    const unsubscribe = useAtlasStore.subscribe(({ cards }) => {
+      const prevCards = prevCardsRef.current;
+
+      const handles = handlesRef.current;
```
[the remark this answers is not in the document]
sagrimson: done

dataweaver/apps/web/src/components/elements/card/chart/data_chart_line.tsx
```diff
@@ -34,10 +34,13 @@ export const DataChartLine = ({ data, width, height }: ChartProps) => {
         tickLine={false}
         axisLine={false}
         tick={{ fontSize: 10, fill: AXIS_COLOR }}
+        tickFormatter={(v) =>
+          Intl.NumberFormat('en', { notation: 'compact' }).format(v)
+        }
```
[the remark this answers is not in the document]
sagrimson: done

dataweaver/apps/web/src/components/elements/card/chart/data_table.tsx
```diff
@@ -5,21 +5,22 @@ interface DataTableProps {
...
-        {data.map(({ year, emissions }) => (
-          <tr key={year}>
-            <td>{year}</td>
-            <td>{emissions}</td>
+        {data.map(({ date, value }) => (
+          <tr key={date}>
+            <td>{date}</td>
+            <td>
+              {Intl.NumberFormat('en', { notation: 'compact' }).format(value)}
+            </td>
```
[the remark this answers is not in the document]
sagrimson: done

dataweaver/apps/web/src/store/store.ts
```diff
@@ -0,0 +1,293 @@
...
+        const { nodes } = get();
+        const chain: HistoryNode[] = [];
+        let currentId: string | null = nodeId;
+
+        while (currentId && chain.length < MAX_ANCESTOR_CHAIN) {
+          const node: HistoryNode | undefined = nodes[currentId];
+          if (!node) break;
+          chain.unshift(node);
+          currentId = node.parentId;
+        }
```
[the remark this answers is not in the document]
sagrimson: done

dataweaver/apps/web/src/store/store.ts
```diff
@@ -0,0 +1,293 @@
...
+          const node = nodes[card.historyNodeId];
+          if (!node) continue;
+          const result = node.results[card.placeDcid];
+          if (!result) continue;
+          for (const entity of result.entities) {
+            dcids.push(entity.dcid);
+          }
+        }
+
+        return dcids;
```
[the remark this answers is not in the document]
sagrimson: done

dataweaver/apps/web/src/server/types.ts
```diff
@@ -5,50 +5,64 @@ export interface ParsedQuery {
...
   name: string;
 }
 
 export interface QueryResult {
   id: string;
-  chartType: ChartType;
   title: string;
   variables: ChartVariable[];
   entities: ChartEntity[];
   metadata: ChartMetadata[];
```
nick-nlb: This isn't part of this PR directly (but does have a bearing on code in the PR that uses this interface):

We are bundling observation/facet/entity information into an interface called `ChartMetadata`, stored in a key labeled `metadata`. 

Most of this information isn't actually metadata, but rather the actual observations themselves (the data/value sequence that makes up the time series), facet/series keys such as measurement method, observationPeriod, unit, etc, (which are a form of metadata, but here are the keys that distinguish a facet).

We will soon have endpoints by available through which Data Weaver will be able to get full, enriched metadata, that would have a matching interface here

For this particular section of the payload, something broader like `data` or `timeSeries` would fit better. Each entry of this array is really a TimeSeries. Most likely it will be rendered as a chart by the frontend (although even now we have the tabular view as well), but here is really a collection of facet-grouped observations.

We can envision how the relationship between the time series and the chart might be more complex when we think about how the frontend might combine time series into charts.

Right now, each chart renders a single time series, but we will soon have single charts that have multiple time series. This could be either a single stat var, multiple entities (GDP, Canada/USA) or a multiple variables, single stat var (GDP/Life Expectancy).

dataweaver/apps/web/src/server/types.ts
```diff
@@ -5,50 +5,64 @@ export interface ParsedQuery {
...
 
-interface ChartEntity {
+export interface ChartEntity {
   dcid: string;
   name: string;
 }
 
 export interface QueryResult {
   id: string;
-  chartType: ChartType;
```
nick-nlb, line 43: Following up on the previous comment about metadata to time series:

I see that the chart type was removed here. We only have line charts right now, but we will need some way for the front end to know "given this particular set of time series, group them together this way, and use this suggested chart type".

We have the array of time series (currently called metadata), but we don't yet have the suggestion of how these end up being grouped together in cards (if at all) and then what the appropriate charting tool is to display them. 

We could have a separate attribute in the QueryResult that represents a suggested grouping of stat vars and entities into individual cards, with each grouping being given a suggested chart type.

dataweaver/apps/web/src/server/types.ts
```diff
@@ -5,50 +5,64 @@ export interface ParsedQuery {
...
   title: string;
   variables: ChartVariable[];
```
nick-nlb: Similarly to the current `ChartMetadata`, the `ChartVariable` and `ChartEntity` aren't really yet specific to charts (although they could be grouped that way later. This is more just a naming things, but these could be just "statVars", "entities", and they could actually be grouped into a metadata interface here (the linking of DCIDs to human-readable names could fit into the broad category of metadata).

dataweaver/apps/web/src/app/api/query/route.ts
```diff
@@ -75,16 +87,21 @@ export async function POST(request: NextRequest) {
...
+        });
 
         // Resolve places
         let places = parsed.places;
-        if (places.length === 0) {
+        if (parsed.isFollowUp && places.length === 0) {
           places =
             selectedEntityDcids.length > 0 ? selectedEntityDcids : [query];
         }
+        if (places.length === 0) places = [query];
```
nick-nlb, line 104: If there is no place, why does the query get used? (If I search for "Tell me about GDP", this will end up being slotted into the `places` section here.

dataweaver/apps/web/src/components/primitives/html_parsed.tsx
```diff
@@ -0,0 +1,71 @@
...
+}) => {
+  const parsedHtml = useMemo(
+    () =>
+      html
+        ? parse(html, {
+            replace: (domNode) => {
+              if (domNode instanceof Element) {
+                for (const attr of Object.keys(domNode.attribs)) {
+                  if (attr.toLowerCase().startsWith('on')) {
+                    delete domNode.attribs[attr];
```
nick-nlb: There is a degree of sanitization happening here, but it could be hardened. Instead of deleting elements that start with "on", could we take a whitelist approach and remove all attributes except for the few that we need (i.e., we may only need class, href, target, rel etc, but not id, name, onX).

We are also dropping a particular set of node types (script, style) but it's hard to be very sure that we have  covered every possible edge case in custom sanitization. I.e., form tags and svgs might be a vector. 

We could keep this sanitization (while hardening it) but if we are handling and passing HTML, we should use a production grade sanitization library.

dataweaver/apps/web/src/components/scopes/atlas/components/in_front_of_canvas/dev_panel.tsx
```diff
@@ -0,0 +1,61 @@
...
+  };
+
+  return (
+    <div className={s.container}>
+      <Button
+        size="small"
+        colorScheme={BUTTON_COLOR_SCHEME}
+        onClick={exportState}
+      >
+        Export State
```
nick-nlb, line 32: This is a very helpful feature! At some point, it might be nice to have this viewable inline (rather than downloading and opening) but for now, this is great.

dataweaver/apps/web/src/server/types.ts
```diff
@@ -5,50 +5,64 @@ export interface ParsedQuery {
...
+export type CardType = 'loading' | 'table' | 'notes' | 'chart';
+
+export interface Insight {
+  title: string;
+  text: string;
+}
 
 export interface CardEntry {
   shapeId: string;
+  historyNodeId: string;
```
nick-nlb, line 32: We are tightly coupling data retrieval (our query history) and presentation (the cards on the canvas and how they decide what data to display). In particular, a card is locked to a single query round and a single entity/variable combination (with the first one displaying by default).

This works for simple tables where users click to spawn individual charts. However, it limits the more complex possibilities that we will need.

For example, f I search for "Life Expectancy and Age Pyramid in North America", I might expect to get multiple stat vars and multiple time series, combined in various ways, and displayed over multiple chart cards, all opened right away. 

In that example, I might get one chart (a line chart) that contains time series for the various countries in NA, as well as a second chart that contains a histogram of the age pyramid (once bar charts are available). 

Relatedly, if we want to manipulate charts directly on the front end (let's say combining two line charts generated over two different queries into one), this tight coupling limits us. Combining two charts on the canvas shouldn't alter or merge the underlying history. 

Thinking aloud about how we could approach this decoupling:

We could introduce a pointer structure in the store:
```
DataSeriesSpec {
  historyNodeId: string;
  placeDcid: string;
  variableDcid: string;
} 
```
Each chart then holds a `series` array of these DataSeriesSpec entries (which would allow us to merge, split, or otherwise customize charts on the canvas by manipulating the series arrays, leaving the underlying query history and data records untouched).

dataweaver/apps/web/src/components/scopes/atlas/query_provider.tsx
```diff
@@ -36,41 +35,166 @@ interface QueryProviderProps {
...
+
+        // Remove the loading placeholder for this place and replace with real cards.
+        const loadingId = `shape:${active.nodeId}__${place}__loading`;
+        const { cardUnregister } = store.getState();
+        cardUnregister(loadingId);
+        active.cardIds = active.cardIds.filter((id) => id !== loadingId);
+
+        // Register table card.
+        const tableId = `shape:${active.nodeId}__${entityDcid}__table`;
+        cardRegister(tableId, active.nodeId, 'table', entityDcid);
```
nick-nlb: Maybe a batch registration of multiple cards for sequences like this?

dataweaver/apps/web/src/store/store.ts
```diff
@@ -0,0 +1,296 @@
...
+            'queryCancel',
+          );
+        },
+
+        querySetProcessing: (val) =>
+          set({ isProcessing: val }, undefined, 'querySetProcessing'),
+        querySetStatus: (val) =>
+          set({ currentStatus: val }, undefined, 'querySetStatus'),
+
+        getAncestorChain: (nodeId) => {
```
nick-nlb, line 249: Can we start to introduce unit tests for these functions (this could be a separate, follow-up PR).

nick-nlb (commented):
Thank you for all this Scott! Looks great, nice to see the store take shape.

A few comments below. Some of them we could roll out into another PR to prevent this one from getting to large/complex (i.e. the rethinking of how query results/cards are connected, unit tests, etc).

For any that you would prefer to address separately, we can place them in the tracker and do them as a follow-up.

dataweaver/apps/web/src/app/api/query/route.ts
sagrimson: This is resolved in the next PR I'm working on that deals with error handling, I can't remember why I put the query in there TBH, I think maybe it was breaking if the array was empty?

dataweaver/apps/web/src/components/primitives/html_parsed.tsx
sagrimson: Yes good call, I added dompurify - much cleaner!

dataweaver/apps/web/src/store/store.ts
sagrimson: For sure, will get a new PR started for this

dataweaver/apps/web/src/server/types.ts
sagrimson: done! Renamed it to 'StatVar' and 'Entity'

dataweaver/apps/web/src/components/scopes/atlas/query_provider.tsx
sagrimson: Ah good idea, done!

dataweaver/apps/web/src/server/types.ts
sagrimson: Added an entry in the tracker for this one :)

dataweaver/apps/web/src/server/types.ts
sagrimson: Renamed to TimeSeries

dataweaver/apps/web/src/server/types.ts
sagrimson: Yes 100%, I was thinking of the history node / card relationship as a one to many, but this seems very obvious now that you mention it 🙈 I added this to the tracker as I think it will be good as a separate PR

dataweaver/apps/web/src/app/api/query/route.ts
sagrimson: (also added to the tracker as 'error handling')

dataweaver/apps/web/src/components/primitives/html_parsed.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const HtmlParsed = ({
+  html,
+  onAction,
+  className,
+}: {
+  html: string;
+  onAction?: (href: string) => void;
+  className?: string;
+}) => {
+  DOMPurify.setConfig({ FORBID_TAGS: ['style'] });
```
nick-nlb: Raw in the function's render execution, this will set the configuration on every render. 

Looking at the `sanitize` docs, it looks like you can pass the configuration directly into sanitize as the second parameter (i.e. inside the useMemo below).

dataweaver/apps/web/src/components/scopes/atlas/sync_store.ts
```diff
@@ -0,0 +1,205 @@
...
+          // Pass raw ID (strip "shape:" prefix) so tldraw shape ID matches store key.
+          const rawId = shapeId.replace(/^shape:/, '');
+          const handle = atlas.add(content, rawId);
+          handles.set(shapeId, handle);
+        }
+
+        // Detect cards whose type changed (loading → real)
+        for (const [shapeId, card] of Object.entries(cards)) {
+          const prevCard = prevCards[shapeId];
+          if (!prevCard || prevCard.type === card.type) continue;
```
nick-nlb, line 168: We can discuss what this functionality should be, but would type be the only differentiator for an existing card that would prompt an overwrite? If I run an import of existing data (maybe from a few turns ago, before I made some changes to a card), I would expect the overwrite to happen to that card even if the type hasn't changed. Let's add to the list to discuss.

nick-nlb (approved):
Thank you for the updates Scott! A couple of comments (one should be quick, and the other one for the list).

dataweaver/apps/web/src/components/primitives/html_parsed.tsx
sagrimson: Ah totally, fixed!

dataweaver/apps/web/src/components/scopes/atlas/sync_store.ts
sagrimson: Yep currently the cards don't change after they've been loaded, but will definitely need to change this logic once we figure out how/when we want to modify cards -- this is also related to decoupling cards from history, so maybe this can be included in that PR, will add a tracker item to discuss!

PR #376, opened 2026-06-23
Implement Follow Up Question And Answer Flow

PauloMFJ:
## Overview

This PR implements follow up question and answer flow. Note: How this is connected is still up in the air so for now just hooked up a very basic flow. **It's very temporary** but will be reworked later so wouldn't pay too much attention to that part for now.

## Screenshot

image

## Testing

* To demonstrate the (mocked up) follow-up flow, click on the fourth sample choice option "What fraction of the world is forest".


nick-nlb (approved):
Thanks Paulo! 

I don't think we've had significant discussion yet on the flow of back-and-forth interaction between DW and the user before DW breaks into showing data. Let's add that to the agenda for Friday.

PR #377, opened 2026-06-23
Implement History Control

PauloMFJ:
## Screenshot

image


nick-nlb (approved):
The buttons themselves look good. They are tied right now to TLDraw's undo/redo functionality (i.e., they undo actions that either the system or user does on the canvas) but of course it will have to be tied carefully into the underlying query/chart state once these two streams are merged. Once you we are able to manipulate charts within cards, it may be possible that some actions that would be eligible for the undo buffer would not TLDraw actions at all.

PR #378, opened 2026-06-23
Implement Chart Style Options Menu

PauloMFJ:
## Overview

Adds the chart-style options menu to chart cards, reachable from the card's action bar.

### What's new

- MenuChartOptions — a radio menu (bar vertical / bar horizontal / line) with Cancel / Update, opened via the chart-options action and dismissible by Escape, backdrop, or Cancel.
- New icons: bar_chart_vertical, bar_chart_horizontal, line_graph_double; renamed bar_chart → bar_chart_outlined and line_graph → line_graph_single.

### Refactor

- Each card variant (CardChart, CardText) now owns its own Card.Base — actions, content, footer, and (for charts) the menu — instead of the atlas shape assembling them. The shape just selects the variant and passes in id + content/state. Promotes - Card.Footer to a shared element.
- The chart-options action and isStyleMenuOpen live together in CardChart, so the button toggles the menu directly.
- Extended how HtmlParsed worked;
   - Updated internal action links to use buttons.
   - Introduced explicit allowedTags list.

## Screenshot

image


dataweaver/apps/web/src/components/elements/card/chart/chart.tsx
```diff
@@ -18,54 +30,125 @@ export interface ChartDatum {
...
+  const editor = useEditor();
+
+  const { open: openExport } = useExportActions();
+  const { runPrompt } = useQueryActions();
+
+  // TODO: Support the different chart styles (for now we always show bar chart)
+  const [selectedStyle, setSelectedStyle] =
+    useState<ChartStyle>('bar-vertical');
+  const [isStyleMenuOpen, setIsStyleMenuOpen] = useState(false);
+
```
nick-nlb, line 74: If you open up the chart menu and click within the chart, it dismisses it (catching the click within the card but outside the menu).

However, if you click anywhere outside the card, the card becomes deselected but the menu remains. 

Noticing the deselection happens, we could fix this in a quick way by adding an effect that catches the change of selection:

```
useEffect(() => {
    if (selection === 'none') {
      setIsStyleMenuOpen(false);
    }
  }, [selection]);
```

It does create a bespoke "chain reaction" to solve the issue.

A more generic clean option could be to use the `useClickOutside` hook that we have in the `use_click_outside` (and we cwuldn't then need the backdrop, although it could still serve a visual purpose over the card.

nick-nlb (changes requested):

dataweaver/apps/web/src/components/elements/card/chart/chart.tsx
PauloMFJ: Good spot! Solved with suggested click outside hook. Much cleaner + allowed me to drop Biome ignore comment!

PauloMFJ:
@nick-nlb Note: As part of merge I also reworked a bit of how `HtmlParsed` was used to clean it up a tad to closer match how I've reworked card as part of this PR.

nick-nlb (approved):
Thank you!

PR #380, opened 2026-06-25
remove 2px padding from top & bottom of notice container

sagrimson:
## Overview

Removes padding from intro card footer

From tracker:
"Intro card footer - The footer in the intro card seems too high for what it is. Let's make it more tight by removing 2 px from the top and bottom"


nick-nlb (approved):
Thank you!

Ultimately these pixel-level tightenings will need to be solidified into the implementation of the tracker item "Review app wide use of spacings (introduce tokenised version / REM sizings)". One aspect of that approach is to establish a spacing scheme where we have a base width (like MUI's 8px) and use multipliers rather than explicit pixel values. (i.e. 1 instead of 8px).

PR #382, opened 2026-06-26
Scott/disambiguation

sagrimson:
## Overview

Adds a disambiguation/follow-up flow for broad or exploratory queries. When a user asks something general (e.g. "tell me about Seychelles"), the AI now returns a conversational summary of what it knows along with suggested analytical directions, rendered as a dedicated FollowUp component instead of immediately generating cards. This lets users narrow their focus before the app commits to a specific data visualization.

Also renames `followUps` → `relatedQueries` on cards to distinguish per-card drill-down prompts from the new top-level disambiguation feature, and removes loading placeholder cards that are no longer needed since results with a follow-up skip card registration entirely.

## Changes Made

- [x] Extended the `data_discovery` skill prompt with a `followUp` JSON schema field and rules for when to include/omit disambiguation
- [x] Added `FollowUp` interface to server types (`summary`, `question`, `options`) and wired it into `QueryResult`
- [x] Updated `PageHome` to derive follow-up state from the latest node's results and render the `FollowUp` component when present
- [x] Refactored `FollowUp` component to accept the new `FollowUpData` type and the original prompt, removing the old `QuestionAndAnswers` interface
- [x] Skipped card registration in `query_provider` when a result contains a `followUp`, so the disambiguation UI takes precedence
- [x] Removed loading placeholder card logic (register on `parsedQuery`, unregister on result/skip/complete) — no longer needed
- [x] Renamed `followUps` → `relatedQueries` / `followUp` → `relatedQuery` on card shapes and helpers to avoid naming collision with the new feature
- [x] Used Zustand selectors in `PageHome` to avoid unnecessary re-renders
- [x] Deleted unused `mock_follow_up.ts` config file


dataweaver/apps/web/src/components/scopes/atlas/query_provider.tsx
```diff
@@ -98,6 +99,14 @@ export const QueryProvider = ({ children }: QueryProviderProps) => {
         cardUnregister(loadingId);
         active.cardIds = active.cardIds.filter((id) => id !== loadingId);
 
+        if (result.disambiguation) {
+          // Handle disambiguation if present
+          queryFail(active.nodeId);
+          querySetProcessing(false);
+          querySetStatus(STATUS.idle);
+          break;
+        }
```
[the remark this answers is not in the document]
sagrimson: done

dataweaver/apps/web/src/components/scopes/page_home/page_home.tsx
```diff
@@ -32,7 +32,29 @@ export const PageHome = () => {
...
+    const disambiguation = disambiguations[0];
+    if (disambiguation && currentStatus === STATUS.complete) {
+      setFollowUp({
+        question: node.query,
+        answer: disambiguation.summary,
+        followUp: disambiguation.question,
+        prompts: disambiguation.options,
+      });
+    }
+  }, [currentStatus, latestNodeId, nodes]);
```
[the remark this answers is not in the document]
sagrimson: done

dataweaver/apps/web/src/components/scopes/atlas/query_provider.tsx
```diff
@@ -99,6 +99,11 @@ export const QueryProvider = ({ children }: QueryProviderProps) => {
         cardUnregister(loadingId);
         active.cardIds = active.cardIds.filter((id) => id !== loadingId);
 
+        if (result.disambiguation) {
+          // Handle disambiguation if present
```
nick-nlb: Can we flesh out in the comment what the purpose of this break is, i.e. "Handle disambiguation" really means skipping the registration of the cards.

dataweaver/apps/web/src/server/types.ts
```diff
@@ -50,6 +51,12 @@ export interface Entity {
   name: string;
 }
 
+export interface Disambiguation {
```
nick-nlb: In a sense, `Disambiguation` is accurate: we are hitting this path when the system decided that it did not have enough information to present the user with data. However, as we move more of what are currently considered "Errors" that are handled by the toasts into follow-up that would be handled here, could we rename this to `followUp`?

As an example, when the user asks a nonsensical question, tries to jailbreak, or asks a question out of the domain of the system, where currently we are toasting errors, we would give more conversational feedback here (and in the case of jailbreaking attempts, vague feedback). The follow-up may or may not contain options, but this would still be a logical vehicle to route it through.

dataweaver/apps/web/src/components/scopes/atlas/query_provider.tsx
```diff
@@ -99,6 +99,11 @@ export const QueryProvider = ({ children }: QueryProviderProps) => {
...
+          break;
```
nick-nlb: nit (can be ignored): Looking at the "break" here, we could turn this into a `return`. It makes it a bit easier to step through the function, as you know right away that when we hit here we are done, rather than needing to look further down the function to see if anything follows.

dataweaver/apps/web/src/components/scopes/page_home/page_home.tsx
```diff
@@ -37,6 +37,21 @@ export const PageHome = () => {
...
     currentStatus !== STATUS.idle;
 
+  useEffect(() => {
+    const node = latestNodeId ? nodes[latestNodeId] : null;
+    const disambiguations = node?.results
+      ? Object.values(node.results)
+          .map((r) => r.disambiguation)
+          .filter(Boolean)
+      : [];
+    const disambiguation = disambiguations[0];
```
nick-nlb: We might have to discuss what the behavior should be if there are multiple disambiguations/follow-ups in a single query, but in this case additional disambiguations are dropped. If we can only have a single follow-up per query round (that might be reasonable), then that should be enforced upstream, resulting in only a single disambiguation to start.

dataweaver/apps/web/src/components/scopes/page_home/follow_up.tsx
```diff
@@ -31,21 +22,22 @@ export const FollowUp = ({ followUp, onSelect }: FollowUpProps) => {
...
       <div className={s['inner-container']}>
-        <p className={s.question}>{followUp.question}</p>
+        <p className={s.question}>{prompt}</p>
 
-        <div className={s.answer}>{followUp.answer}</div>
+        <div className={s.answer}>{followUp.summary}</div>
+        <div className={s.answer}>{followUp.question}</div>
 
-        {followUp.prompts.length > 0 && (
+        {followUp.options.length > 0 && (
```
nick-nlb: Short of a more rigorous type/formatting check of the response, could we add optional chaining here to ensure that it is an array (i.e. it has a length)?

This ultimately comes from an unsanitized LLM generated structure on the server side. We should also add rigorous checks on the server side for any data structures generated by the model to catch any deviations of structure upstream.

dataweaver/apps/web/src/app/api/query/route.ts
nick-nlb, line 197: If we don't receive any variables, we short out of the response here (skipping the disambiguation step). Especially as we recast more of what we are now considering "errors" to "followup", it becomes more and more likely that a very ambiguous response wouldn't contain any variables. In that case, we would still want to disambiguate. 

If we consider a query "What about France" (or even something more specific, but not "data-ish" like "What is the capital of France", in either case, we would not expect to get stat vars back. 

We actually do get stat vars back in that case, which I think is incorrect. The agent logic tries very hard to come up with something. However, we should assume that this isn't always the case. 

Relating also to the fact that right now we always do get back something, if I type in "What's the capital of France", it does correctly ask me for some follow-up (i.e. it has recognized that my query doesn't point to any specific data it can retrieve) but it still returns in the payload query results and data relating to generic stat vars. There might be exceptions, but generally this sort of query response should not have data.

dataweaver/apps/web/src/components/scopes/page_home/page_home.tsx
```diff
@@ -37,6 +37,21 @@ export const PageHome = () => {
...
+    if (node && disambiguation && currentStatus === STATUS.complete) {
+      setFollowUp(disambiguation);
+    } else {
+      setFollowUp(null);
+    }
+  }, [currentStatus, latestNodeId, nodes]);
```
nick-nlb: To limit the effect's dependency scope, since we actually only want this to trigger when the latest node changes, at the top where we pull out all the nodes from the store, we could more selectively pull out only the latest node. The effect's triggering would then be limited to only the one node that affects it.

dataweaver/apps/web/src/components/scopes/atlas/query_provider.tsx
nick-nlb: Now that we have a more complex flow that involves a back-and-forth conversation with the agent (prompt / followUp / ...), the loading state being set here is hasty. 

We set up a loading state for a card before we really know that a card will be placed up there. 

Can we defer the appearance of the loading card until later in the process (when we know that we aren't going back to the followUp loop).

dataweaver/apps/web/src/components/scopes/page_home/page_home.tsx
sagrimson: Ah yes I meant to add a TODO here, definitely need to figure out what to do here. We get one disambiguation per place currently if you ask something like 'Tell me about France & Brazil'.

dataweaver/apps/web/src/components/scopes/atlas/query_provider.tsx
sagrimson: We were actually talking about this in our meeting this morning :) 

This makes the loading card unnecessary since we don't know if we're going to need a disambiguation or not until the full results return. I was worried this would make it feel less responsive because the cards take a while to show up, but with the status component popping up right away and streaming the current status it actually still feels good to me. LMK what you think!

dataweaver/apps/web/src/components/scopes/atlas/query_provider.tsx
sagrimson: Yes, done!

dataweaver/apps/web/src/server/types.ts
sagrimson: Yeah I like that, plus it makes the naming consistent between the component & the data :)
There was already a 'followUps' array for suggested follow-up questions that show up in cards, but I changed that to 'relatedQueries' to distinguish them from the follow-up flow.

dataweaver/apps/web/src/components/scopes/atlas/query_provider.tsx
sagrimson: done!

dataweaver/apps/web/src/components/scopes/page_home/page_home.tsx
sagrimson: Added the TODO, let's discuss in our meeting

dataweaver/apps/web/src/components/scopes/page_home/follow_up.tsx
sagrimson: Totally, good catch, done!

dataweaver/apps/web/src/components/scopes/page_home/page_home.tsx
sagrimson: Ah yeah thanks for this, using zustand selectors here now and it's much better!

dataweaver/apps/web/src/app/api/query/route.ts
sagrimson: Oh yeah you're right, I missed that! There will probably be followups for most of these, I just added a check for && !followUp here -- also added a check right below this for cases where there is no entityDcId

dataweaver/apps/web/src/components/scopes/atlas/query_provider.tsx
nick-nlb: I was also mulling the question of whether or not the card skeleton was necessary now (it's effectively a proxy for a loading spinner that may or may not turn into a card). Let's discuss in the meeting, but as you say we could potentially remove it!

sagrimson: I ended up removing it -- definitely easy to add it back in there if there's a need, but it honestly feels cleaner to me this way. You can test it out now on the staging link https://dataweaver.datacommons.foo/

dataweaver/apps/web/src/components/scopes/page_home/page_home.tsx
nick-nlb: We did touch on this in the meeting, but I think it would be worth digging deeper into how we really want this to behave (as solving this properly might require us to rework some of the query/response flow here). Given the "back-and-forth" aspect of the followUp mechanism, it might make more sense to have only one possible query result per round that has a followUp. Follow-up would then be a result type in its own right, a first-class memberof the history node (rather than being coupled to an extracted place). I.e. if I search for "Tell me something", there is no place to extract, but the feedback should be something like "Please ask me a more specific question about data".

Any one conversation round could give us:
- zero or more sets of board level cards
- zero or one followUps
- both board level cards and a single follow-up
- never no cards or followUp

The processing path could then be:
- we try to parse the query for places/query parts. 
- if one of the places/query does not trigger disambiguation, then we send it back as a query result and continue checking the places
- if a place/query does trigger disabiguation, we put it aside
- at the end, if we have any disambiguation needs, we do it all at once, and send back a followUp message (a single one for the whole conversation round)
- Likewise, if the query was adversarial or otherwise irrelevant/unparseable, we send back a single followUp message saying so.
- Regardless, if there is follow-up it will only come a single time, near the end of the event stream

nick-nlb (approved):
Thank you Scott!

PR #388, opened 2026-07-01
Overhaul Card Placement System

PauloMFJ:
## Card placement

- Cards now place into a responsive grid: 3 columns on laptop+, 2 on tablet, 1 on mobile,
  driven by the shared breakpoint tokens with a per-breakpoint gutter (`CARD_GRID` in config).
- Placement flows like text: each card lands against the previous card's right edge, wrapping
  after the column count to a new row below the tallest card of the row above.
- Placement is camera-independent — breakpoints match the window in screen px and card widths
  are capped per breakpoint, so panning/zooming never changes where or how big cards land.
- Moving a card re-roots the grid there: the next card (new or pasted) lands beside it and
  rows wrap back to its x. Detected structurally (x/y-only user change), so nudges count too.
- Copy/paste and duplicates flow onto the grid like new cards; alt-drag duplicates stay where
  the pointer drops them.
- tldraw's paste re-centering is vetoed — pasted cards keep their grid position instead of
  jumping to the viewport center when they land off-screen.
- The placement cursor survives deletion (pruned per card) and remounts (re-derived from the
  cards on canvas), so undo/redo and reloads self-heal.

## Camera

- `keepInView` now frames the grid being built (cards placed since the last re-root) instead
  of the whole canvas, so a new grid far from older cards doesn't zoom out to fit everything.
- Mobile reads as a vertical feed: the camera only pans (never zooms out to fit), and a batch
  of cards arriving at once pans to the first of the set instead of chasing each one.

## Structure

- Split canvas side effects into one module per concern: `register_card_placement`,
  `register_card_clones` (mirroring + cascade deletes), `register_card_store_sync`,
  alongside the existing `register_card_tab_navigation`.
- `AtlasProvider` slimmed to theme setup, four registrations, and the `add/update/remove`
  API; clone tracking moved out of the component into the clones module.
- Added `scopes/atlas/PLACEMENT.md` documenting the placement model, camera rules, and
  trade-offs; linked from `AGENTS.md`.
- `contentToShape` takes an explicit size; new `CardShape` type alias shared across modules.


nick-nlb (approved):
Thank you Paulo. The placement is definitely improved.

Let's get this in and up on foo, where everyone can experiment and give feedback on specific edge cases.

PR #389, opened 2026-07-01
Update Card Text HtmlParsed Styles

PauloMFJ:
## Screenshots

Before:
image

After:
image

### DIFF
--- a/dataweaver/apps/web/src/styles/core/_reset.scss
+++ b/dataweaver/apps/web/src/styles/core/_reset.scss
@@ -85,7 +85,6 @@
 
   table {
     border-spacing: unset;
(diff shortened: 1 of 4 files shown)


dataweaver/apps/web/src/styles/core/_reset.scss
```diff
@@ -85,7 +85,6 @@
 
   table {
     border-spacing: unset;
-    border-collapse: collapse;
```
nick-nlb, line 88: We don't have any other affected tables, but I think collapsing the border by default is a reasonable reset state (in which case we would apply the reversal of the collapse in the text.module.scss).

nick-nlb (approved):

PR #391, opened 2026-07-03
Update status messages

sagrimson:
## Overview

Updates the status messages for STATUS.checkingSafety & STATUS.parsingQuery to be a bit less specific.


nick-nlb (approved):
Thank you!

PR #363, opened 2026-06-11
feat(narratives): PR 1 - Core Shell, Routing, and Branding

ddebasmita-lab:
This Pull Request introduces the initial structure and core layout shell for the new Narratives custom web interface under the /narratives folder. This is the first PR in a phased sequence to set up the Custom Data Commons narratives tool.

Key Additions:
Project Scaffolding: Configured React + TypeScript + Vite setup under narratives/ with strict type checking enabled ("strict": true in tsconfig.json).
Google Development Standards: Followed Google HTML/CSS and TypeScript styling guidelines (including block-wrapped control statements, clean import aliasing, and standard self-closing tags).
Branding & Design System: Integrated a BrandingProvider to dynamically fetch and inject instance-specific customization (e.g. brand-primary colors, brand font families, custom logos) as CSS variables.
Layout & Navigation: Implemented the base application shell (App.tsx), responsive sidebar/header components, and a lightweight hash-based router (useHashRoute).
Git Protection: Configured project-specific .gitignore and .dockerignore files to prevent committing local terraform caches, .tfvars config files, and generated build folders.

Files Included in this Phase:
narratives/tsconfig.json & vite.config.ts
narratives/index.html & package.json
Core shell components: Header.tsx, Sidebar.tsx, App.tsx
Configs and hooks: navConfig.ts, useHashRoute.ts, BrandingContext.tsx, useBranding.ts

PR #392, opened 2026-07-05
feat(narratives): PR 2 - print optimization, naming standardizations,…

ddebasmita-lab:
## Overview

Second PR for the **narratives** Data Agent UI. PR 1 (#363) landed the shell —
routing, branding, and the streaming scaffold. This PR builds the actual chat
experience on top of it: multi-session conversations, real chart rendering via
Data Commons web components, a config-driven Key Metrics dashboard, and
single-answer print/PDF export. It also folds in the code-review feedback from
this PR's own earlier rounds (naming standardizations and print hardening).

Scope is the `narratives/` frontend only (TypeScript + React 19 + Vite +
Tailwind v4). ~42 files, no backend changes.

## Background / motivation

PR 1 rendered a single ephemeral conversation with placeholder charts. To make
the Data Agent usable we needed: conversations that survive navigation and
refresh; charts that show correct per-place data with real source attribution;
a metrics landing page an instance owner can configure without code; and a way
to export an answer for sharing. Each of those is one section below.

## Changes made

**Multi-session chat (persisted).** Conversations are stored in `localStorage`
and exposed through a context provider, so turns survive SPA navigation and
browser refresh. A session drawer lists past chats with new/switch/delete;
empty "New chat" drafts are collapsed so the list stays tidy. Desktop shows a
left-rail list; mobile uses a two-level slide-in panel.
_Files: `hooks/chat_session_context.tsx`, `components/drawer_session.tsx`,
`components/sidebar.tsx`._

**Data Commons web-component charts.** Replaced the custom Recharts layer with
official `<datacommons-*>` components, wrapped by `DataCommonsComponent`. Each
component owns its own data fetch, source caption, "About this data" / "API
code" footer, and metadata modal, wired to the actual DCIDs the agent returned.
This fixes the prior bug where charts with missing `placeDcids` defaulted to
country/USA.
_Files: `utils/datacommons_component.tsx`, `components/tile_chart.tsx`._

**Config-driven Key Metrics dashboard.** The metrics page renders tiles from a
metrics config rather than hardcoded charts, so an instance can re-skin it
without touching components.
_Files: `components/metrics_page.tsx`, `types/metrics.ts`._

**Print / PDF export.** Export one Q&A panel via a scoped `@media print`
strategy that hides everything outside the selected subtree. Hardened against
detached targets and concurrent print runs, with cleanup in `finally`.
_Files: `utils/print_element.ts`, `components/button_export_pdf.tsx`, `index.css`._

**Streaming, follow-ups, citations.** SSE streaming with a stop control and a
single-stream guard (no concurrent connections), agent follow-up questions, and
inline `[N]` citation chips.
_Files: `hooks/use_sse_chat.ts`, `components/data_agent.tsx`,
`components/questions_follow_up.tsx`, `components/chip_citation.tsx`._


narratives/src/api/dc_observations.ts
```diff
@@ -41,21 +38,19 @@ export interface PointResponse {
...
   date: string;
-  [key: string]: string | number | undefined;
-}
+};
 
 /**
- * Represents the page's host origin, dynamically set to window.location.origin
- * in the browser, or fallback empty string in node/test environments.
+ * Base origin for API URLs — the current page's origin in the browser
+ * (the instance proxies /api itself), empty during SSR/tests.
```
juliawu: "SSR/tests" is confusing because it looks like a filepath. How about we use "empty during server-side rendering (SSR) or testing."

narratives/src/index.css
```diff
@@ -5,39 +5,50 @@
...
  *
  * Names mirror the Figma GM3 token vocabulary where possible.
+ *
+ * BRAND-DRIVEN TOKENS: tokens below written as `var(--brand-*, <fallback>)` are
+ * recolored at runtime from the instance's branding.json. hooks/use_branding.ts
+ * fetches the config bucket and sets `--brand-primary`, `--brand-accent`,
+ * `--brand-text`, `--brand-text-muted`, `--brand-surface`, `--brand-background`,
+ * `--brand-font`, `--brand-font-body` on :root. The `<fallback>` after the comma
+ * is the shipped design value used when no branding.json is present — keep it in
+ * sync with DEFAULT_BRAND in useBranding.ts. The hero-gradient stops are
```
juliawu: useBranding.ts -> use_branding.ts

narratives/src/index.css
```diff
@@ -192,24 +209,131 @@ body {
...
+.dc-chart-card datacommons-slider {
+  display: block;
+  width: 100%;
+  max-width: 100%;
+  box-sizing: border-box;
+}
+
+/* ────────────────────────────────────────────────────────────────────
+ * Print stylesheet: when the Export PDF action fires (see
+ * utils/printElement.ts), `body[data-print-target]` is set and the
```
juliawu: printElement.ts -> print_element.ts

narratives/src/index.css
```diff
@@ -192,24 +209,131 @@ body {
...
-  body[data-print-target] nav,
-  body[data-print-target] [data-non-print="true"] {
-    display: none !important;
-  }
   body[data-print-target] {
     background: #ffffff !important;
   }
-  body[data-print-target] #answer-panel {
+  /* Hide every body descendant that is NOT on the path to the active panel
+     and NOT inside the active subtree. The JS in utils/printElement.ts has
```
juliawu: printElement.ts -> print_element.ts

narratives/src/utils/print_element.ts
```diff
@@ -0,0 +1,56 @@
...
+ * index.css hide everything else on the page (visibility: hidden across the
+ * document, visibility: visible on the active subtree only). After the print
+ * dialog closes the tags are removed so the page returns to normal.
+ *
+ * This replaces the older "hardcoded #answer-panel" strategy that printed
+ * EVERY answer panel in the chat — there are now many answer panels in a
+ * conversation, each must be exportable in isolation.
+ */
+export function printElement(target: HTMLElement | null) {
+  if (!target) return;
```
juliawu: You might also want to check that the target is connected to the DOM (and hasn't been disconnected by a concurrent react re-render) here
```typescript
if (!target || !target.isConnected) return;
```

narratives/src/utils/print_element.ts
```diff
@@ -0,0 +1,56 @@
...
+  // observer pick up the new width synchronously.
+  const onBeforePrint = () => {
+    window.dispatchEvent(new Event("resize"));
+  };
+  window.addEventListener("beforeprint", onBeforePrint);
+
+  // Defer so any state-driven class updates (e.g. focus rings) flush first,
+  // then trigger the platform print dialog.
+  requestAnimationFrame(() => {
+    window.print();
```
juliawu: I would wrap the `window.print()` in a try/catch/finally block, that way we make sure the cleanup of tags will run even if the print throws an exception.

narratives/src/utils/print_element.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  // When the browser switches to print layout the DC web components'
+  // internal ResizeObserver needs to fire so their SVGs redraw at the
+  // narrower print-page width instead of staying at the on-screen
+  // measurement. `beforeprint` runs AFTER the print layout is committed
+  // but BEFORE the page is snapshot; dispatching a resize here lets the
+  // observer pick up the new width synchronously.
+  const onBeforePrint = () => {
+    window.dispatchEvent(new Event("resize"));
+  };
```
juliawu, line 55: Should the cleanup steps also include a resize event so that the SVGs reset to the original on-screen measurement?

narratives/src/utils/print_element.ts
juliawu: Furthermore, what happens when the print button is double clicked or the user clicks the print button in two separate panels in quick succession?

Here's the scenario I'm worried about: Imagine the user (or test agent) triggers this printElement function for two targets, A and B, in quick succession. If this function is not atomic, it's possible that both target A and target B's ancestors all get tagged with printAncestor=true before `requestAnimationFrame` fires. Then, when the first `requestAnimationFrame` runs, you end up with a print containing both A and B elements.

You might also want to check if `document.body.dataset.printTarget === "true"` and return if so, because that means a print sequence is already in progress.
```typescript
if (!target || !target.isConnected || document.body.dataset.printTarget === "true") return; 
```

narratives/src/utils/dc_component.tsx
```diff
@@ -0,0 +1,112 @@
...
+  | "datacommons-map"
+  | "datacommons-pie"
+  | "datacommons-gauge"
+  | "datacommons-highlight"
+  | "datacommons-ranking"
+  | "datacommons-scatter"
+  | "datacommons-slider";
+
+/** Attribute bag for one <datacommons-*> element; values are stringified into HTML attributes. */
+export type DCAttrs = Record<string, string | number | boolean | undefined>;
```
juliawu: Let's be more explicit with names. How about something like:

`DCTag` -> `DataCommonsComponentTagName`
`DCAttrs` -> `DataCommonsComponentAttributes`
`DC` -> `DataCommonsComponent`

and rename the file to `datacommons_component.tsx`. While more verbose, the longer names are self-documenting and play better with IntelliSense/Autocomplete IDE tools.

narratives/src/components/chip_citation.tsx
```diff
@@ -36,18 +37,16 @@ export function ChipCitation({ n }: { n: number }) {
...
- * Walks a children array and replaces `[N]` patterns inside strings with
- * <ChipCitation /> elements. Returns a new children array suitable for
+ * Walk a children array and replace `[N]` patterns inside strings with
+ * <CitationChip /> elements. Returns a new children array suitable for
  * React rendering. Pass any react-markdown component's `children` through
  * this helper to apply citation styling consistently inside p, li, td,
  * strong, em, etc.
  */
-export function withChipCitations(
+export function withCitationChips(
```
juliawu: The current naming makes the usage of withCitationChips unclear. In react higher-order functions prefixed with `with` like `withTheme` or `withRouter` typically wrap other `React.ComponentType` components and are called outside the render loop. It looks like this function is more of a helper function for parsing/formating. How about renaming to something like `renderWithCitations`

narratives/src/components/list_sources.tsx
```diff
@@ -52,23 +53,31 @@ export function ListSources({ sources }: ListSourcesProps) {
           color: COLOR_BODY,
         }}
       >
-        {sources.map((source, index) => {
-          const position = index + 1;
+        {sources.map((s, i) => {
```
juliawu: I'd still prefer the explicit `source, index` instead of `s, i` here.

narratives/src/components/tile_chart.tsx
```diff
@@ -1,645 +1,143 @@
...
-        </ResponsiveContainer>
-      );
-    }
-
+/**
+ * Maps an agent vizType to a DC web component tag. Anything without an
+ * explicit mapping falls back to line, which is the most forgiving
+ * renderer (single variable + single place is always valid).
+ */
+function dcTagFor(viz: ChartItem["vizType"]): DCTag {
```
juliawu: Let's keep with the verb-led naming convention and use something like `getDcComponentTag` here and `getDcComponentAttributes` below for clarity.

narratives/src/hooks/use_branding.ts
```diff
@@ -121,36 +238,67 @@ export function useBranding(): { branding: Branding; loaded: boolean; error: str
...
         });
         if (!brandResp.ok) {
           throw new Error(`/agent/brand HTTP ${brandResp.status}`);
         }
-        const rawConfig = (await brandResp.json()) as { brand_config_url?: string };
-        const { brandConfigUrl }: BrandConfigResponse = {
-          brandConfigUrl: rawConfig.brand_config_url,
-        };
-        if (!brandConfigUrl) {
+        const { brand_config_url, branding: directBranding } =
```
juliawu: Could we make `brand_config_url` camelCase instead, so `brandConfigUrl`? This would better match typescript naming convention.

narratives/src/hooks/chat_session_context.tsx
```diff
@@ -66,13 +378,31 @@ export function ChatSessionProvider({ children }: { children: ReactNode }) {
...
+ * Light helper for the drawer UI. Exported so the drawer can reuse the same
+ * relative-time format if we want to tweak it in one place.
+ */
+export function formatRelativeTime(ts: number): string {
+  const diff = Date.now() - ts;
+  if (diff < 60_000) return "just now";
+  if (diff < 3_600_000) return `${Math.floor(diff / 60_000)} min ago`;
+  if (diff < 86_400_000) return `${Math.floor(diff / 3_600_000)} hr ago`;
+  if (diff < 172_800_000) return "yesterday";
+  const days = Math.floor(diff / 86_400_000);
```
juliawu: what do you think of defining constants like
```typescript
const MS_PER_MINUTE = 60_000;
const MS_PER_HOUR = 60 * MS_PER_MINUTE;
const MS_PER_DAY = 24 * MS_PER_HOUR;
```
so that you can then do
```typescript
export function formatRelativeTime(ts: number): string {
  const diff = Date.now() - ts;
  if (diff < MS_PER_MINUTE) return "just now";
  if (diff < MS_PER_HOUR) return `${Math.floor(diff / MS_PER_MINUTE)} min ago`;
  if (diff < MS_PER_DAY) return `${Math.floor(diff / MS_PER_HOUR)} hr ago`;
  if (diff < 2 * MS_PER_DAY) return "yesterday";
  const days = Math.floor(diff / MS_PER_DAY);
  if (days < 7) return `${days} days ago`;
  return new Date(ts).toLocaleDateString("en-US", {
    month: "short",
    day: "numeric",
  });
}
```
This would make the different cases and intent more clear at a glance.

narratives/src/components/drawer_session.tsx
```diff
@@ -1,8 +1,321 @@
...
+  onDelete,
+}: Readonly<{
+  sessions: ReturnType<typeof useChatSession>["sessions"];
+  currentSessionId: string;
+  onSelect: (id: string) => void;
+  onDelete: (id: string) => void;
+}>) {
+  return (
+    <ul className="flex-1 overflow-y-auto no-scrollbar list-none p-0 m-0">
+      {sessions.map((s) => {
```
juliawu: This map is long enough where I'd prefer we rename `s` to `session` to be explicit.

narratives/src/components/data_agent.tsx
```diff
@@ -1,12 +1,265 @@
...
+  // turn is added (e.g. follow-up question click) — not on every re-render
+  // caused by streaming text chunks arriving.
+  const prevTurnsLength = useRef(turns.length);
+
+  // `override` lets a caller (e.g. a suggestion chip in InitialView) submit
+  // a message directly without first round-tripping through the `query`
+  // state — setState is async, so chaining setQuery → handleSend in one tick
+  // would otherwise send an empty string.
+  const handleSend = async (override?: string) => {
+    const m = (override ?? query).trim();
```
juliawu: using `message` instead of `m` would help readability here

juliawu (commented):
Thanks for the new additions! I see a lot of refactoring, especially in making changes compatible with the updates made in the last PR -- great job! I left some comments that should be quick fixes.

Could you also fill out the PR description? It'll be really helpful down the line to have that context.

narratives/src/api/dc_observations.ts
```diff
@@ -120,8 +111,9 @@ export async function fetchPoint(
...
 
 /**
- * Pivots the time-series API response into one row per date, with each variable as a
- * column. Variables that don't have a value on a given date are omitted
+ * Pivots a series response into the time-series chart shape (line/bar over
+ * time, single place per variable): one row per date, with each variable as
+ * a column. Variables that don't have a value on a given date are omitted
  * from that row.
  */
 export function pivotSeriesToRows(
```
nick-nlb, line 119: These functions (pivotSeriesToRows, pivotPointsToRowsByPlace, etc) are good candidates for unit testing.

narratives/src/app.tsx
```diff
@@ -1,53 +1,74 @@
 /**
```
nick-nlb, line 1: `App.tsx` -> `app.tsx`

narratives/src/components/icons.tsx
```diff
@@ -0,0 +1,70 @@
+/**
```
nick-nlb, line 1: File name: `icons.tsx`

narratives/src/components/sidebar.tsx
```diff
@@ -32,16 +34,17 @@ export function Sidebar() {
...
+          {/* Start a fresh chat thread — does NOT delete other sessions; they
+              remain accessible from the drawer above. */}
           <button
             type="button"
             aria-label="Start new chat"
             onClick={newSession}
             className="w-12 h-12 flex items-center justify-center rounded-full hover:bg-button-hover transition-colors text-on-surface-variant"
           >
-            <SquarePen size={20} strokeWidth={1.5} />
+            <NewChatIcon size={22} />
```
nick-nlb, line 47: Here we are explicitly setting the size of the icon directly in the consuming component. This is a magic number, and will have the effect of making standardizing (and adjusting) the icons difficult. 

There are some different options for handling this, but I would lean towards having size be of type "sm" | "md" | "lg", and having the pixel size of these defined on the theme level (with "md" being 22, "lg" 24 for example.

narratives/src/components/data_agent.tsx
```diff
@@ -1,12 +1,265 @@
...
+          style={{
+            background:
+              "linear-gradient(to top, #FFFFFF 0%, rgba(255,255,255,0) 100%)",
+          }}
+        />
+      </div>
+
+      {/* Follow-up input — flush against the fade (no top margin) so the
+          gradient reads as a continuous merge rather than a separator. */}
+      <div className="w-full flex flex-col items-center gap-3 bg-white pb-4 sm:pb-6">
```
nick-nlb, line 153: I've noticed that there are a substantial number of hard-coded color references in the app (`bg-white`, `border-gray-200`). These should all be derived from the theme.

This might be a sizable job in itself, and can be set aside for a subsequent PR, but can we make sure we have the task lined up to make sure that all items that might reasonably vary from theme-to-theme be derived from theme tokens rather than being hard-coded?

narratives/src/components/data_agent.tsx
```diff
@@ -1,12 +1,265 @@
...
+        thoughts={turn.thoughts}
+        streaming={isStreaming && turn.status !== "done" && turn.status !== "error"}
+        done={turn.status === "done"}
+        status={turn.status}
+      />
+
+      {/* Skeleton while the turn is streaming but no synthesis text has
+          arrived yet. Covers the full pre-text window: mcp (tools running),
+          kb (knowledge base lookup), and synthesis (model thinking before
+          first token). Previously this only checked `mcp`, leaving a ~3s
```
nick-nlb: "Previously this only checked `mcp`, leaving a ~3s blank gap between mcp_complete and the first text chunk."

This last sentence in the comment is on the change made (and so, what the code used to do), rather than what the code does.  While it can be interesting and temporarily useful, discussions on fixed errors within the code itself distracts from the functional description. Can we remove this, as well as similar comments above: 
```
Showing both was redundant ("Querying data tools…"
          duplicated "Searching Data Commons for relevant indicators…")

 Phase loader removed — ReasoningBlock below is the sole progress
          indicator. Showing both was redundant ("Querying data tools…"
          duplicated "Searching Data Commons for relevant indicators…")
```

If the comment on a change is to a reviewer, it could go in the PR itself.

narratives/src/components/metrics_page.tsx
```diff
@@ -1,12 +1,317 @@
...
+/**
+ * Key Metrics Dashboard.
+ *
+ * CONFIG-DRIVEN COMPOSITION: the tab+tile layout comes from
+ * `branding.metrics.tabs` fetched at mount. When absent, DEFAULT_METRICS_TABS
+ * below is used. This preserves the "configuration over images" property — a
+ * new state instance changes its dashboards by editing branding.json in its
+ * config bucket, no image rebuild needed.
+ *
+ * IMPORTANT — why we use dangerouslySetInnerHTML for the DC components:
```
nick-nlb: This comment explains why `dangerouslySetInnerHTML` is used - but it doesn't look like it is used anymore.

narratives/src/components/icons.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+interface IconProps {
+  readonly size?: number;
+  readonly className?: string;
+}
+
+/**
+ * `pen_spark` — New chat.
+ */
+export function NewChatIcon({ size = 22, className = "" }: IconProps) {
```
nick-nlb, line 21: Following up on that previous comment about icon sizing:

We could requisition this `size` prop to standardize the sizing across the application (i.e. have it take "sm", "md", "lg"). 

We could make this easier here by having a non-exported base Icon component that takes the size, looks up the pixel value that represents in the theme, and uses that pixel value.

The other components would then wrap this new base Icon component.

Another thing we could do, to make these components as flexible as possible, is to have them take SVG props and spread them in (something we do in base DC - although we could do it a bit differently here).

We could use a hook to handle the conversion of common props (like size) into something the icon would understand:

```
// Ultimately we would want these to come right out of the theme (in the branding context?)
const ICON_SIZES = {
  sm: 20,
  md: 22,
  lg: 24,
} as const;

export type IconSizeToken = keyof typeof ICON_SIZES;

export interface IconProps extends SVGProps<SVGSVGElement> {
  size?: IconSizeToken;
}
export function getSvgProps(size: IconSizeToken = "md", props: SVGProps<SVGSVGElement> = {}) {
  const dimension = ICON_SIZES[size];
  return {
    width: dimension,
    height: dimension,
    fill: "currentColor",
    "aria-hidden": true,
    ...props,
    className: `shrink-0 select-none ${props.className ?? ""}`.trim(),
  };
}
```

And then one of the actual icons:

```
export function SomeIcon({ size = "lg", ...props }: IconProps) {
  return (
    <svg
      viewBox="0 -960 960 960"
      {...getSvgProps(size, props)}
    >
      <path d="..."/>
    </svg>
  );
}
```

I played around and tested with the above and it seems to work, but haven't fully vetted it. Feel use any of this.

narratives/src/components/block_reasoning.tsx
```diff
@@ -1,8 +1,16 @@
 /**
```
nick-nlb, line 1: Following from the category-first naming strategy generally employed in the app, this would be `block_reasoning.tsx`

narratives/src/components/block_reasoning.tsx
```diff
@@ -226,6 +220,7 @@ function SparkleIcon() {
   );
 }
 
+/** Chevron icon for the reasoning header; rotates 180° when collapsed. */
 function ChevronDown({ rotated }: { rotated?: boolean }) {
```
nick-nlb, line 224: These should be rolled out into the Icons library.

But importantly as we do that, the icons should not be responsible for their rotation - i.e., ChevronDown is not a special case of "an icon that can be rotated".  Any icon should be rotatable (and that would be controlled from outside of the icon).

Once we have moved our icons to accept SVG props (see the earlier comments), we will be in a position to do that, where we could rotate the icon with something like:

```
<ChevronDownIcon 
  className={`transition-transform duration-150 ${open ? "rotate-0" : "-rotate-90"}`} 
/>
```

narratives/src/components/button_export_pdf.tsx
nick-nlb, line 77: This should also be rolled into the icon library. While doing that - since this is Material icon from the same set as the other icons, it could be updated to have the same viewbox (and of course, with the paths updated to fit).

nick-nlb (commented):
Thank you Debasmita! A few comments added in below.

narratives/src/api/dc_observations.ts
ddebasmita-lab: will add in the next PR

narratives/src/components/sidebar.tsx
ddebasmita-lab: will cover this and the color tokens in the next PR

narratives/src/components/button_export_pdf.tsx
ddebasmita-lab: will be covered in next PR

narratives/src/components/block_reasoning.tsx
ddebasmita-lab: will be addressed in next PR

narratives/src/components/icons.tsx
ddebasmita-lab: Will be addressed in the next PR

nick-nlb (approved):
Thank you for all the updates Debasmita (and noting the deferrals to a subsequent PR for some of the changes - I think that's a good idea in the interest of keeping the PRs small where possible).

PR #394, opened 2026-07-06
Fix Menu's Appearing Behind Home Content

PauloMFJ:
## Overview

Implemented `Portal` element to fix open `Menu` showing behind home content.

Fixed focus events in `InFrontOfCanvas` getting consumed by tldraw. To do so implemented new `capture` config option on `useKeyDown`.

Re-ordered zoom options:
image

Tickets:
image
image


dataweaver/apps/web/src/components/primitives/portal.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * The primitive is unopinionated: it applies no styling of its own.
+ */
+export const Portal = ({ className, children }: PortalProps) => {
+  const [mountedElement, setMountedElement] = useState<HTMLDivElement | null>(
+    null,
+  );
+
+  useEffect(() => {
+    const element = document.createElement('div');
+    if (className) element.className = className;
```
nick-nlb: This effect really does two things. Assuming that the className might change independently of the initial mounting, this triggers the mounting (ie.. setMountedElement) when the class changes. We can separate these into two separate effects, a primary mounting effect and a secondary className change effect.

dataweaver/apps/web/src/hooks/use_keydown.ts
```diff
@@ -123,14 +152,17 @@ export const useKeydown = (
...
       };
 
       const listener: Listener = { keyCode, callback };
-      addListener(eventTarget, listener, eventHandler);
-      return () => removeListener(eventTarget, listener);
+      addListener(eventTarget, listener, eventHandler, capture);
+      return () => removeListener(eventTarget, listener, capture);
     }
-  }, [config.isEnabled, config.target, keyCode, callback]);
+  }, [config.isEnabled, config.target, config.capture, keyCode, callback]);
```
nick-nlb, line 167: If the callback is not memoized coming in to this, then we might see a significant amount of churn here. This isn't new to this PR, but something to flag for later.

dataweaver/apps/web/src/components/elements/menu.tsx
```diff
@@ -12,41 +12,54 @@ interface MenuProps
...
+  prefersMotion,
+  children,
+  onClose,
+  ...rest
+}: MenuWithPrefersMotionProps) => {
+  const containerRef = useRef<HTMLDialogElement>(null);
+
   useKeydown('Escape', onClose);
 
-  // TODO: For now this doesn't seem to really work due to TLDraw consuming
```
nick-nlb, line 22: This comment also exists in `menu_chart_options.tsx` and can be removed from there as well.

dataweaver/apps/web/src/hooks/use_keydown.ts
```diff
@@ -31,7 +41,14 @@ interface Query {
   listeners: Set<Listener>;
 }
 
-const QUERIES = new Map<EventTarget, Query>();
+// Capture-phase and bubble-phase listeners require separate `addEventListener`
+// registrations on the same target, so each phase gets its own registry.
+const BUBBLE_QUERIES = new Map<EventTarget, Query>();
```
nick-nlb: Since we are using DOM elements as keys, let's update this to a WeakMap (it shouldn't matter in this particular case, but generally, it will allow a safety-net for garbage collection where destroyed DOM elements can be garbage-collected and removed from the map automatically).

nick-nlb (approved):
Thank you Paulo - a couple of comments before merging, but otherwise looks great.

dataweaver/apps/web/src/components/elements/menu.tsx
PauloMFJ: This fix is unrelated the the Menu Chart Options issue here which I fixed in: https://github.com/datacommonsorg/tools/pull/399. Will leave as is as the latter PR removes comment alongside the corresponding fix.

dataweaver/apps/web/src/hooks/use_keydown.ts
PauloMFJ: Good spot. In theory with React Compiler functions are memorised so shouldn't be an issue? previously experienced some weird instances where React Compiler didn't behave as expected and lack of memorised callback caused issues. Added ticket to review later.

dataweaver/apps/web/src/components/primitives/portal.tsx
PauloMFJ: Good spot - updated!

PR #398, opened 2026-07-06
Prevent Prompt From Collapsing

PauloMFJ:
Ticket:
image


nick-nlb (approved):
Thank you Paulo - this makes a big difference.

PR #399, opened 2026-07-06
Improve Canvas Tab Support

PauloMFJ:
## Overview

Adds keyboard tab navigation through card content on the tldraw canvas, which tldraw normally swallows to cycle shape selection.

## Changes

- **New `registerCardTabNavigation`** — a capture-phase Tab/Shift+Tab handler that steps through a flat list of every card's focusable elements in reading order (top-to-bottom, left-to-right), wrapping at the ends. It skips culled cards, stays out of the way of open dialogs and tldraw text editing, dives into the selected card when tabbing in from the canvas, and calls `markEventAsHandled` so tldraw ignores the event.
- **Wired into `AtlasProvider`** — registers on editor mount, cleans up on unmount.
- **Card CSS** — reveal the action bar on `:focus-within` (not just selection) so it's reachable via Tab.
- **`useFocusTrap` improvements** — export and broaden `TABBABLE_ELEMENTS` (add inputs/selects/textareas, exclude disabled) for reuse; guard against non-`HTMLElement` active elements.
- **Re-enabled focus trap** in the chart options menu — removed the stale TODO now that Tab handling works.
- **CSS layer reorder** — `base` now sits after `primitive`. This was changed because `tldraw` has a default `outline: none` on `*` elements which was overriding our `base` style. Our `base` is quite lean so this change is safe.


dataweaver/apps/web/src/components/elements/card/chart/menu_chart_options.tsx
```diff
@@ -60,9 +60,6 @@ export const MenuChartOptions = ({
 
   useKeydown('Escape', onClose);
 
-  // TODO: For now this doesn't seem to really work due to TLDraw consuming
```
nick-nlb, line 63: I made a comment about this in a previous PR (which is now out-of-date due to this change).

dataweaver/apps/web/src/components/scopes/atlas/register_card_tab_navigation.ts
```diff
@@ -0,0 +1,156 @@
...
+      return;
+    }
+
+    // Only handle tab events from the canvas itself (e.g. right after clicking
+    // a card to select it) or from inside a card
+    if (activeElement !== container && !activeElement.closest(CARD_SELECTOR)) {
+      return;
+    }
+
+    const focusables = getCardFocusables(editor);
```
nick-nlb: This runs on every tab hit. This means that as a user tabs through the canvas, with each key press the focusables array is rebuild (despite it only changing if the canvas actually changes).

It's not critical at this stage, but this could get costly on a very large canvas with many cards. Can we create a tracker item to update this to cache it (and invalidate the cache whenever the canvas changes)?

dataweaver/apps/web/src/styles/layers.css
```diff
@@ -1 +1 @@
-@layer reset, root, base, primitive;
+@layer reset, root, primitive, base;
```
nick-nlb, line 1: There's content in the FRONTEND.md documentation that refers to the order and still documents it the other way around.

nick-nlb (approved):
The functionality over all works well.

However, if you have a card selected, and start tabbing between cards, the subsequent cards where the focus now falls does not get selection (i.e. the original selected card remains the selected one, and for a time both cards have an action bar at the top (until focus leaves the action bar). 

This in itself might be correct behavior, but it brings up a question that we should put up for discussion: the keyboard navigation is powerful, but you don't have a way to select or deselect cards. 

Otherwise a minor comment and a task that we could either do in this PR or move to the tracker.

dataweaver/apps/web/src/components/elements/card/chart/menu_chart_options.tsx
PauloMFJ: Replied there + kept here!

dataweaver/apps/web/src/components/scopes/atlas/register_card_tab_navigation.ts
PauloMFJ: Good spot! Updated.

PauloMFJ:
> The functionality over all works well.
> 
> However, if you have a card selected, and start tabbing between cards, the subsequent cards where the focus now falls does not get selection (i.e. the original selected card remains the selected one, and for a time both cards have an action bar at the top (until focus leaves the action bar).
> 
> This in itself might be correct behavior, but it brings up a question that we should put up for discussion: the keyboard navigation is powerful, but you don't have a way to select or deselect cards.
> 
> Otherwise a minor comment and a task that we could either do in this PR or move to the tracker.

Opened a ticket for this - would be good to discuss!

PR #401, opened 2026-07-07
Implement Updated Prompt Button Tone

PauloMFJ:
## Overview

Reworks button tones to add an explicit **active** state and simplifies the prompt submit button's visual behavior.

## Changes
- **Button component**: Added an `isActive` prop (default `false`) that sets a `data-is-active` attribute; removed the unused `prominent` tone.
- **Button styles**: Split hover vs. active states — hover only applies when not active, and added dedicated active-state base/content/border styling per tone.
- **Design tokens**: Dropped the `button-prominent-*` tokens, added `*-active` color tokens for all remaining tones, and introduced a new `accent-strong` color (used for accent hover).
- **Canvas tools**: Tool buttons now use the `subtle` tone with `isActive` instead of swapping to the `accent` tone.
- **Prompt submit button**: Switches variant/tone based on whether there's input — `flat`/`accent` when filled, `border`/`subtle` when empty (replacing the removed `prominent` tone).

Video:
https://github.com/user-attachments/assets/aee55409-7046-4cc9-8093-674893af4682

Ticket:
image


nick-nlb (approved):
Thank you!

PR #403, opened 2026-07-07
Implement Card Resize + Scroll Rework

PauloMFJ:
## Overview

Cards can now be manually resized on the canvas, and their content scrolls internally under a pinned title instead of resizing the whole card.

### Changes

- **Manual card resizing** — Enabled tldraw resizing on the card shape (`onResize` via `resizeBox`) with a `CARD_SIZE_MIN` of 300×220. Added an `isManuallyResized` prop that, once set, tells `useCardAutoHeight` to stop syncing height so the user's chosen size sticks.
- **Cursor-only resize UI** — Reworked the selection foreground overlay to keep hit-testing/resize-cursors active while rendering no handles or selection frame.
- **New `Card.Content` component** — A titled scroll region that fills the card: the title/header is pinned to the top and the body scrolls beneath it, with a divider that appears on scroll (`data-has-scrolled`). Wheel-capture logic moved here so cards scroll from anywhere over the card, not just the scroll area.
- **Auto-height rework** — `useCardAutoHeight` now measures container + content refs (accounting for fixed chrome like title/padding) instead of `scrollHeight`.
- **Responsive charts** — Line chart now uses recharts `ResponsiveContainer` (width fills the resizable card; height stays fixed), dropping the hardcoded `CHART_WIDTH`.
- **Config rename** — `CARD_VARIANT_MAX` → `CARD_VARIANT_SIZE_DEFAULT` to reflect that it's now a default (resizable) footprint rather than a hard max; updated all references.
- **Base card cleanup** — `CardBase` now takes a `childrenContainerRef` prop; removed auto-height/scroll wiring from it.
- **Text card styles** — Migrated markdown spacing to `:not(:first-child)` top-margins. Ensures that we don't add margin bottom or top on first / last elements.

## Screenshots:

New card with scroll:
image


dataweaver/apps/web/src/components/elements/card/content.tsx
```diff
@@ -0,0 +1,60 @@
...
+export const CardContent = ({
+  contentContainerRef,
+  title,
+  children,
+}: CardContentProps) => {
+  const contentOuterContainerRef = useRef<HTMLDivElement>(null);
+
+  const [hasScrolled, setHasScrolled] = useState(false);
+
+  const getCachedCanScroll = useCachedResizeValues((element: HTMLElement) => {
```
nick-nlb: We are getting a stale cache here, where the resize listener isn't triggering and the cached value is stale.

The symptoms of this are that when you scroll inside the card, and then resize the card, the cache doesn't recognize the change in the cards scrollability state. If you resize the window (the resize event is attached to the window) it will.

We could fix this and make it simpler by making use of the existing ResizeObserver and managing a local state for tracking scrollability.

We then no longer need the window-level cached values and the hook.

Suggestion, something like this (something I did locally that works):

```
const [canScroll, setCanScroll] = useState(false);
useEffect(() => {
  const outer = contentOuterContainerRef.current;
  if (!outer) return;
  const observer = new ResizeObserver(() => {
    setCanScroll(outer.scrollHeight > outer.clientHeight);
  });
  observer.observe(outer);
  return () => observer.disconnect();
}, []);
```

nick-nlb (changes requested):

dataweaver/apps/web/src/components/elements/card/content.tsx
PauloMFJ: Good feedback! I ended up combining previous version with new local resize hook. Wanted to avoid triggering reflow events (`setCanScroll(outer.scrollHeight > outer.clientHeight)`) before user actually attempts to scroll.

nick-nlb (approved):
Thank you!

PR #406, opened 2026-07-15
Extend How Border Radius Is Shown Across App

PauloMFJ:
### Prompt and Intro

Before:
image

After:
image

### Status

Before:
image

After:
image

### Follow Up

Before:
image

After:
image

### Menu, Toast and Cards

Before:
image

After:
image

## Ticket
image


miss-o-soup:
I might be wrong, but looking at the screenshots, it seems the that larger radius is only applied to the top left corner. 
It should be to the top left and the bottom right (see attached image) - only for the special cards, the ones in the canvas stay as they are at 16. 
Also the input box we should leave it as it was uniform (all corners the same radious)
RoundedCorners

PauloMFJ:
> I might be wrong, but looking at the screenshots, it seems the that larger radius is only applied to the top left corner. It should be to the top left and the bottom right (see attached image) - only for the special cards, the ones in the canvas stay as they are at 16. Also the input box we should leave it as it was uniform (all corners the same radious) 

Gotcha! I missed that it was top left and bottom right. Will fix.

Also will revert prompt change - what about toasts? Should that match as it was or be included in change?

miss-o-soup:
Thank you. Let's say toasts are not special so we keep them uniform.

PauloMFJ:
@miss-o-soup have updated the screenshots ^ let me know how things look?

nick-nlb (approved):

PR #409, opened 2026-07-16
restyle facet dropdown to match figma

sagrimson:
## Overview

- Updates the styles for the facet selector in chart cards.
- adds tokens for facet selector colors

Screenshot 2026-07-16 at 3 23 25 PM

Screenshot 2026-07-16 at 3 23 19 PM

### DIFF
--- a/dataweaver/apps/web/src/components/elements/card/chart/facet_selector.module.scss
+++ b/dataweaver/apps/web/src/components/elements/card/chart/facet_selector.module.scss
@@ -1,9 +1,10 @@
 .container {
   position: relative;
-  display: flex;
+  display: grid;
   flex-shrink: 0;
   gap: 8px;
   margin-block: 18px;
+  color: rgb(var(--color-facet-selector-color));
   background: rgb(var(--color-facet-selector-surface));
   border-radius: var(--border-radius-extra-small);
   box-shadow: 0 0 0 1px rgb(var(--color-facet-selector-border));
@@ -12,15 +13,14 @@
 .prefix {
   @include type-body-small;
 
-  position: absolute;
-  top: 0;
-  left: 0;
   display: flex;
   flex-shrink: 0;
+  grid-area: 1 / 1;
   gap: 0;
+  place-self: start;
   align-items: flex-start;
   font-weight: 500;
-  color: rgb(var(--color-card-content-muted));
+  color: rgb(var(--color-facet-selector-prefix));
   background: rgb(var(--color-facet-selector-surface));
   box-shadow: 0 0 0 3px rgb(var(--color-facet-selector-surface));
   transform: translate(16px, -50%);
@@ -31,14 +31,12 @@
   @include type-body-medium;
 
   display: flex;
+  grid-area: 1 / 1;
   gap: 16px;
   align-items: center;
   justify-content: space-between;
   width: 100%;
   padding: 12px 16px;
-  color: rgb(var(--color-facet-selector-color));
-  background: none;
-  border: none;
 }
 
 .trigger {
@@ -56,7 +54,7 @@
   position: absolute;
   top: 100%;
   left: 0;
-  z-index: 100;
+  z-index: $z-index-above-content;
   width: 100%;
   min-width: 280px;
   margin-top: 5px;
@@ -80,12 +78,12 @@
   background: transparent;
 
   &[data-is-selected="true"] {
-    background: rgb(var(--color-button-accent-subtle-base));
+    background: rgb(var(--color-facet-selector-option-selected));
   }
 
   &:not([data-is-selected="true"]) {
     @include hover {
-      background: rgb(var(--color-button-accent-subtle-base-hover));
+      background: rgb(var(--color-facet-selector-option-hover));
     }
   }
 }
--- a/dataweaver/apps/web/src/components/elements/card/chart/facet_selector.module.scss
+++ b/dataweaver/apps/web/src/components/elements/card/chart/facet_selector.module.scss
@@ -3,43 +3,51 @@
   display: flex;
   flex-shrink: 0;
   gap: 8px;
-  margin-block: 12px;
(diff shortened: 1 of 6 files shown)


dataweaver/apps/web/src/components/elements/card/chart/facet_selector.tsx
```diff
@@ -28,14 +28,15 @@ export const FacetSelector = ({
   if (!currentFacet || facets.length <= 1) {
     return currentFacet ? (
       <div className={clsx(s.label, s.container)}>
```
nick-nlb: If there is only one facet, this combines label and container together on the same container, and so the container gets the transparent background intended for the label (rather than the container).

dataweaver/apps/web/src/components/elements/card/chart/facet_selector.module.scss
```diff
@@ -51,49 +59,33 @@
...
 
   &[data-is-selected="true"] {
-    background: rgb(var(--color-control-accent) / 12%);
+    background: rgb(var(--color-button-accent-subtle-base));
   }
 
   &:not([data-is-selected="true"]) {
     @include hover {
-      background: rgb(var(--color-control-accent) / 6%);
+      background: rgb(var(--color-button-accent-subtle-base-hover));
```
nick-nlb: Here we are using the button token directly in another component (see the line in AGENTS.md).

dataweaver/apps/web/src/components/elements/card/chart/facet_selector.module.scss
```diff
@@ -51,49 +59,33 @@
   z-index: 100;
   width: 100%;
   min-width: 280px;
-  margin-top: 4px;
+  margin-top: 5px;
   overflow: hidden;
-  background: rgb(var(--color-card-surface));
-  border-radius: var(--border-radius-small);
+  background: rgb(var(--color-facet-selector-surface));
```
nick-nlb: While this component is a facet selector, it feels like the underlying component is a custom select box that could be used more generally in other contexts.

What do you think of converting this into a generic Select (with the styling applied to it), and then FacetSelector would be a wrapping component that uses Select internally?

dataweaver/packages/tokens/src/colors.json
```diff
@@ -104,6 +104,10 @@
   "card-action-content-hover": "$accent",
   "card-accent-subtle": "$accent-subtle",
 
+  "facet-selector-surface": "$card-surface",
+  "facet-selector-border": [196, 199, 197],
+  "facet-selector-color": "$card-content",
```
nick-nlb: The token here is aliasing another components tokens.

dataweaver/apps/web/src/components/elements/card/chart/facet_selector.module.scss
```diff
@@ -3,43 +3,51 @@
...
 
-  color: rgb(var(--color-card-content-muted));
+  display: flex;
+  gap: 16px;
+  align-items: center;
+  justify-content: space-between;
+  width: 100%;
+  padding: 12px 16px;
+  color: rgb(var(--color-facet-selector-color));
+  background: none;
```
nick-nlb: This is worth double-checking (i.e. are these already reset this way upstream), but are these restates of already existing resets? (see the frontend style guide about restating the reset)

nick-nlb (changes requested):
It looks good. Some comments (most small, but the biggest one, besides cross-component token usage, is about the possibility of a more generic select (with generic select tokens), that is worth discussion with Paulo, etc. Let me know what you think.

Another consideration is that when we have a single facet (and therefore an "inert" select box, it still looks like the select box should be clickable. This might be a discussion for design, but I think this should look and feel more "readonly" in this case.

dataweaver/apps/web/src/components/elements/card/chart/facet_selector.tsx
sagrimson: Ah very true, thanks!

dataweaver/apps/web/src/components/elements/card/chart/facet_selector.module.scss
sagrimson: fixed!

dataweaver/apps/web/src/components/elements/card/chart/facet_selector.module.scss
sagrimson: Totally agree, done!

dataweaver/packages/tokens/src/colors.json
sagrimson: fixed!

dataweaver/apps/web/src/components/elements/card/chart/facet_selector.module.scss
sagrimson: done!

sagrimson:
> It looks good. Some comments (most small, but the biggest one, besides cross-component token usage, is about the possibility of a more generic select (with generic select tokens), that is worth discussion with Paulo, etc. Let me know what you think.
> 
> Another consideration is that when we have a single facet (and therefore an "inert" select box, it still looks like the select box should be clickable. This might be a discussion for design, but I think this should look and feel more "readonly" in this case.

Thanks Nick, all feedback addressed! Totally agree on the select box, also sorry about all the token stuff, a little bit of a RTFM issue on my part 😳

I'll run the design question past Matt & see what he thinks. I agree it does seem like the single facet could use something to differentiate from the interactive dropdown

miss-o-soup:
I was not able to check on the facet new additions, but what I was hoping to have is the text centered with the facet info, and when there are many more facets to have a chevron (with the proper hover over behaviour and all that).
34cpUE5puLH4XV4

nick-nlb (approved):
Thank you for the follow-up Scott!

PR #410, opened 2026-07-20
refactor(narratives): icon tokens, theme colors, pivot tests (PR 2 follow-up)

ddebasmita-lab:
## Overview

Addresses the review items from PR 2 that were explicitly deferred to a
follow-up. Three cohesive pieces: an icon-size token system, a theme-token
migration for hard-coded colors, and unit tests for the observation helpers.

## Changes made

**Icon system**
- `icons.tsx`: add semantic size tokens (`xs`/`sm`/`md`/`lg`) with pixel values
  in one map, a `getSvgProps` base helper, and `IconProps extends SVGProps` so
  every icon accepts standard SVG attributes.
- Call sites now pass size tokens (`size="lg"`) instead of magic numbers.
- Rolled the previously-inline `ChevronDown` (reasoning header) and the
  Export-PDF icon into the icons library as `ChevronDownIcon` / `ExportIcon`.
- Rotation is now caller-controlled — the chevron rotates via `className`
  (`-rotate-90` when collapsed); the icon is no longer aware of rotation.

**Theme colors**
- Replaced hard-coded Tailwind colors (`bg-white`, `border-gray-*`, `red-*`,
  the `black/30` scrim) with semantic tokens defined once in `index.css`
  `@theme`: `surface`, `surface-muted`, `outline`, `outline-variant`,
  `divider`, `error` / `error-strong` / `error-surface`, and `scrim`. Values
  match the previous palette, so there is no visual change — but an instance
  can now re-skin from one place.

**Tests**
- `dc_observations.test.ts`: unit tests for `pivotSeriesToRows`,
  `pivotPointToRowsByPlace`, `pickSourceFromFacets`, `getPrettyPlaceName`, and
  `getPrettyVariableName`; removed the stale "add tests" TODO.

## Testing done

- [x] `npm run lint` (`tsc --noEmit`, strict) — clean.
- [x] `npm run build` (vite) — clean; verified the new token utilities compile.
- [x] `npm test` (vitest) — 19/19 (was 4).

### DIFF
--- a/narratives/src/components/icons.tsx
+++ b/narratives/src/components/icons.tsx
@@ -1,70 +1,103 @@
 /**
- * @fileoverview Provides shared inline SVG icon components used across the UI.
+ * @fileoverview Shared inline SVG icon components and their size-token system.
  */
+
+import type { SVGProps } from "react";
 
 /**
- * Google Material Symbols icons used across the chat UI.
+ * Icon pixel sizes keyed by a semantic token. Centralised here so icon sizing
+ * is consistent and adjustable in one place, rather than hard-coded as magic
+ * numbers at each call site.
  *
- * These are the real `pen_spark` (New chat), `notes_spark` (Chats) and
- * `menu` (accordion / hamburger) glyphs. They use the Material Symbols
- * 0 -960 960 960 viewBox and fill `currentColor` so they inherit text color.
+ * TODO(theme): ideally these come from the branding/theme layer so a downstream
+ * instance can rescale icons; kept local until the theme exposes a size scale.
  */
+const ICON_SIZES = {
+  xs: 16,
+  sm: 20,
+  md: 22,
+  lg: 24,
+} as const;
 
-interface IconProps {
-  readonly size?: number;
-  readonly className?: string;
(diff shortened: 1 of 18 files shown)


narratives/src/components/icons.tsx
```diff
@@ -1,70 +1,103 @@
...
   );
 }
 
 /**
- * `menu` — hamburger / accordion toggle.
+ * `edit` (pencil) — leading glyph on the Export PDF button. Kept on its native
+ * 24×24 viewBox; matches the Figma leading-icon slot (node 3427-16783).
  */
-export function MenuIcon({ size = 24, className = "" }: IconProps) {
+export function ExportIcon({ size = "sm", ...props }: IconProps) {
```
juliawu, line 97: [Suggestion] For this and other icons in the file, name based on what is drawn by the svg instead of how it's used (e.g. instead of `ExportIcon` use `PencilIcon`. ). This would allow the icons to be reused elsewhere in the UI in the future. For example, if we decide we also want an "edit" button with the pencil icon as well, it's a little strange to import "ExportIcon" to use in the button.

juliawu (approved):
Thank you for updates! This looks good to me. I have just one comment that doesn't need to block this PR, but consider adding it in a followup.

PR #411, opened 2026-07-20
Implement Designed Import / Export Flow

PauloMFJ:
## Changes

- New import flow inside `in_front_of_canvas` - a menu with file browse plus drag-and-drop upload (`data-is-dragging-over` state, blocks canvas gestures leaking through to tldraw) that restores the workspace from an exported JSON file.
- Update export flow with a full-canvas export radio option (`EXPORT_OPTION_CANVAS`) that's always offered regardless of selection and pairs with the new import menu.
- Update serialization with addition of `STATE_VERSION` and a typed `ImportError` (`malformed` | `version-mismatch` + `fileVersion`), driving full error-toast messaging when an import fails.
- Drop legacy `DevPanel` and its `use_dev_mode` hook, replaced by the import / export flow.
- Extract `Card` component from `StatusEmpty` - existing component was getting quite large as it was.
- Add new `import` icon.
- Update `control-divider` token to use existing `$surface-muted` for consistency.

## Screenshots

image
image
image
image


dataweaver/apps/web/src/store/serialization.ts
```diff
@@ -36,33 +50,52 @@ export const exportState = (): void => {
...
+      ? parsed.version
+      : undefined;
+
+  // Catch version mismatch
+  if (parsedVersion !== STATE_VERSION) {
+    throw new ImportError('version-mismatch', parsedVersion);
+  }
+
+  if (!hasValidState(parsed)) {
+    throw new ImportError('malformed');
```
beets: Consider adding different messages for these errors so they are easier to debug.

dataweaver/apps/web/src/store/serialization.ts
```diff
@@ -36,33 +50,52 @@ export const exportState = (): void => {
...
+  }
+
+  const parsedVersion =
+    'version' in parsed && typeof parsed.version === 'number'
+      ? parsed.version
+      : undefined;
+
+  // Catch version mismatch
+  if (parsedVersion !== STATE_VERSION) {
```
beets, line 74: Undefined versions also end up triggering this error, although it's not quite a version-mismatch but an invalid Json. Let's add a separate error message for that.

beets (approved):
Thanks for the change - this looks nice! Some suggested changes below but approving early to unblock.

dataweaver/apps/web/src/store/serialization.ts
PauloMFJ: Good call! Updated :)

dataweaver/apps/web/src/store/serialization.ts
PauloMFJ: Good catch! Fixed.

PR #414, opened 2026-07-21
Implement Ability To Drag Selection Actions Bar

PauloMFJ:
## Overview

- Enable dragging the whole multi-selection by pressing its blue actions bar, mirroring how a single card's actions bar drags the card.
- Extract the shared drag mechanics (pointer capture, screen→page delta, `updateShapes`, history markers) into a reusable `useShapeDragHandle` hook, and reduce `useCardDragHandle` to a thin wrapper over it so the two paths no longer duplicate logic.
- Move `Selection` into its own `selection/` folder and split it into `useSelectionPosition` (presence + box placement) and `useSelectionDragHandle` hooks, matching the card base's hook-per-concern structure.
- Updated use of `cursor: grabbing` to use `cursor: all-scroll` to closer match tldraw behaviour.

## Ticket

image


beets (approved):
Nice changes!

I've noticed that drag doesn't update the z-index of the card which seems counterintuitive to me. Here's an example (it also appears with a single card drag). Something we can discuss separately and doesn't block this PR.

Image

PauloMFJ:
> Nice changes!
> 
> I've noticed that drag doesn't update the z-index of the card which seems counterintuitive to me. Here's an example (it also appears with a single card drag). Something we can discuss separately and doesn't block this PR.
> 
> Image

Good spot, and I agree - will address in a separate PR.

PR #424, opened 2026-07-23
Fix Dragging Scrollbar Causing Jump

PauloMFJ:


beets (approved):
nice find! 

could you merge after applying the gemini suggestion?

PR #425, opened 2026-07-24
feat(narratives): stopped-turn UX, reasoning polish, token-usage readout

ddebasmita-lab:
## Overview

Follow-up UI chunk for the narratives Data Agent, bringing the remaining
front-end refinements into the repo. No backend/build-config changes.

## Changes made

- **Stopped-turn UX**: when the user clicks Stop mid-stream, the half-finished
  reasoning/answer is replaced with a "Stopped per your request" note instead
  of leaving truncated chrome (`data_agent.tsx`, `block_reasoning.tsx`).
- **Reasoning-aside polish**: the reasoning section reads as a muted aside with
  a left bar; heading styling and spacing refined (`block_reasoning.tsx`).
- **Mobile drawer nav** refinements to the session drawer (`drawer_session.tsx`).
- **Token-usage readout (debug-only)**: a `Tokens IN / OUT / Total` block in the
  header, shown **only** when the page is opened with `?debug=tokens`. It reads
  an optional `usage` SSE event; off by default and invisible to normal users.
  Temporary instrumentation for gauging per-query cost (`header.tsx`,
  `use_sse_chat.ts`).

## Testing done

- [x] `npm run lint` (`tsc --noEmit`, strict) — clean.
- [x] `npm run build` (vite) — clean.
- [x] `npm test` (vitest) — passing.

### DIFF
--- a/narratives/src/components/block_reasoning.tsx
+++ b/narratives/src/components/block_reasoning.tsx
@@ -152,11 +152,10 @@ export function ReasoningBlock({
   const [open, setOpen] = useState(defaultOpen);
   const label = phaseLabel(streaming, status);
 
-  // Flatten to a single markdown string (newline-separated). Phase grouping
-  // existed in the old ThoughtsPanel; the new Figma design renders the
-  // reasoning as one continuous block. When the agent emits no thoughts,
-  // we render a short status placeholder so the section stays present
-  // above the response card per Figma 3427-16715.
+  // Flatten to a single markdown string (newline-separated): the reasoning
+  // renders as one continuous block, not grouped per phase. When the agent
+  // emits no thoughts, we render a short status placeholder so the section
+  // stays present above the response card per Figma 3427-16715.
   const joined =
     thoughts.length > 0
       ? thoughts.map((thought) => thought.text).join("\n\n")
--- a/narratives/src/components/block_reasoning.tsx
+++ b/narratives/src/components/block_reasoning.tsx
@@ -2,7 +2,7 @@
  * @fileoverview Renders the agent's collapsible reasoning/thoughts block for a turn.
  */
 
-import { useState } from "react";
+import { useState, type CSSProperties } from "react";
 import ReactMarkdown from "react-markdown";
 import remarkGfm from "remark-gfm";
 import { ChevronDownIcon } from "./icons";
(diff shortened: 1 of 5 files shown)


narratives/src/components/block_reasoning.tsx
```diff
@@ -127,29 +201,54 @@ export function ReasoningBlock({
...
               ),
               ol: ({ children }) => (
-                <ol className="list-decimal pl-5 mb-2 space-y-0.5">
+                <ol
+                  className="list-decimal space-y-0.5"
+                  style={{ ...BODY_BLOCK_STYLE, paddingLeft: BAR_INSET + 20 }}
+                >
                   {children}
                 </ol>
               ),
```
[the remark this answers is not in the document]
ddebasmita-lab: Good catch — agreed that the inline left-border on every list renderer means nested lists draw a duplicate inner bar.

Deferring this to a focused follow-up: doing it properly means moving the bar/spacing onto CSS direct-child selectors (`.reasoning-markdown > p`, `> ul`, `> ol`) and removing the inline styles (a CSS rule cannot override the inline border otherwise). That is a visual-layout change best verified by eye against nested-list content, so it is safer as its own change that can be QA-checked in the browser.

The other two comments in this review (history filtering for stopped/incomplete turns, and evaluating the `?debug=tokens` flag per render) are addressed in the latest push.

narratives/src/components/header.tsx
```diff
@@ -31,6 +43,20 @@ export function Header() {
...
+      {latestUsage && (
+        <div
+          className="hidden sm:block text-caption text-on-surface-variant leading-tight border border-outline rounded-md px-3 py-1.5 shrink-0"
+          aria-label="Token usage for the latest query"
+        >
+          <div>Tokens IN: {latestUsage.input.toLocaleString()}</div>
+          <div>Tokens OUT: {latestUsage.output.toLocaleString()}</div>
+          <div>Total: {latestUsage.total.toLocaleString()}</div>
+        </div>
+      )}
```
juliawu, line 58: Given that we're planning on extracting the header out to be its own component, should the token-usage readout and its logic be moved somewhere else?

juliawu (commented):
Thanks for the updates! Overall, this looks good to me. Just one question about where we want to place the debug token usage location.

narratives/src/components/header.tsx
ddebasmita-lab: Good point. Since it is temporary `?debug=tokens` instrumentation and the header extraction hasn't landed yet, I've left it in place for now to keep this PR scoped. When the header is pulled out into its own component, the readout (and `isTokenDebugEnabled` / the `usage` reduce) should move with it — or into a small dedicated `TokenUsageReadout` component the header composes, so the header stays presentational. Happy to do that as part of the header-extraction change.

ddebasmita-lab: Also this is were the tokens show when \?debug=tokens. This is as per Adriana's designs.
https://screenshot-v2.corp.google.com/4MSrECe82284KEs

juliawu: Sounds good, LGTM!

juliawu (approved):

PR #439, opened 2026-08-06
feat(narratives): add config contract, static assets and lockfile

ddebasmita-lab:
Makes the narratives app deployable from a clean clone. Three related gaps:

- `package-lock.json` was never committed, so `npm ci` — used by the image build — failed outright. All 25 dependencies float on `^` ranges, so pinning the tree is what makes a clone reproduce the tested build.

- `public/` was missing entirely while four code paths referenced it: `/send.svg` (view_initial, data_agent), `/loader.png` (block_reasoning) and `/logo.png` (header, as a fallback). Those requests 404'd.

- `config/` documents the per-instance contract — the branding and agent schemas plus fillable examples — which the agent and UI read from the instance's config bucket at runtime. Instance-specific values stay out of the repo: copy the `.example.json` files and edit them.

`template_vars` is kept but now says plainly that placeholder rendering is not implemented; the previous wording claimed the agent already substituted `{{instance.*}}`, which it does not.

---


narratives/public/logo.png
juliawu, line 1: Can be changed later, but @miss-o-soup  should we have a different "default" logo in the main repo?

narratives/config/agent-config.schema.json
```diff
@@ -0,0 +1,177 @@
...
+  "description": "Schema for this instance's agent-config.json (lives at the per-instance config bucket root). Mirrors the field surface of upstream config.json.example. Same schema is consumed in dev mode (UI panel) and prod mode (bucket file). One instance, one file \u2014 no overlay model across instances.",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "gemini": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "api_keys": {
+          "description": "Key pool for rotation. SHAPE only \u2014 real values live in Secret Manager (GEMINI_API_KEYS_SECRET env var) in prod mode. The agent's prod-mode loader rejects inline keys with a warning.",
```
juliawu: Let's require Secret Manager instead of inline keys throughout. This makes it safe to check in configs in repos, and defends against inadvertent exposed keys.

juliawu (commented):
Thank you for the cleanup! Just a few comments. I also think the gemini review comments are worth looking at.

juliawu (approved):
Thanks for the cleanup!

PR #441, opened 2026-08-14
Improve the style of the agent thinking card

pablonoel:
## Overview

Improves the style, behaviour and visual aspects of the agent thinking card, the focus flow, the intro and follow up cards. Adds new elements accordingly. 

## Related Issues

Fixes # (issue number)

## Changes Made

- [x] List significant changes
- [x] List UI changes (include screenshots if applicable)

*New: Button huge size*

- button.tsx — size prop adds 'huge' option.
- button.module.scss — huge shares large's pill padding/gap but forces full (9999px) radius; adds bigger label typography (20px/22px, 16px padding-block).

*New: IconStatusIndicator*

- status_indicator.tsx (new file) — small div-based icon component.
- status_indicator.module.scss (new file) — two-layer icon: spinning ring (::before, conic-gradient masked to a circle, animates via prefers-motion) + static sparkle (::after, masked to an SVG path).

*New: pulsing-card effect (shared mixin)*

- _effects.module.scss (new file) — pulsing-background (bottom-anchored gradient glow, $bottom-offset + $is-animated params, respects prefers-motion) and pulsing-card (shared card shell: position/grid/overflow/background/border/radius + the glow).
- includes.scss — forwards the new effects.module.
- status.module.scss — container now uses pulsing-card (animated); padding/gap reduced; .indicator-message becomes a flex row to hold the new icon.
- intro.module.scss — uses pulsing-card(54px, false) (offset glow, no animation).
- follow_up.module.scss — uses pulsing-card($is-animated: false).

*Autofocus behavior*

- use_auto_focus.ts (new file) — useAutoFocus(ref, isEnabled) hook, focuses element when isEnabled flips true.
- page_home.tsx:
- hasFollowUp derived flag extracted; followUp state effect simplified to depend on it.
- promptInputRef created and passed to Prompt; autofocuses the prompt textarea except while Status or FollowUp panel is showing.
- passes isStatusVisible down to Prompt.
- status.tsx — autofocuses the Cancel button on mount; status message now renders IconStatusIndicator + text inside a role="status" live region.

*Prompt panel: disabled state while query runs*

- prompt.tsx — new inputRef and isStatusVisible props; textarea disabled and submit button isDisabled gated on isStatusVisible; content-container gets data-is-status-visible for the dim/opacity style; submit blocked while status visible.
- prompt.module.scss — container radius medium to large; content-container fades to 0.5 opacity when data-is-status-visible="true"; input text now medium-weight (500), fixed 20px size, uses -subtle content color for both text and placeholder (was default/dimmed).

*Button size bumps on existing call sites*

- follow_up.tsx — option buttons medium to large.
- intro.tsx — example-prompt buttons medium to huge.

https://github.com/user-attachments/assets/78cb41db-7b04-4f63-b412-0613dac341f4

## Testing Done
Describe the steps you took to test these changes (please also list commands ran if possible).

- [x] Unit tests passed
- [x] Integration tests passed
- [x] Manual verification

## Checklist

- [x] I have followed the style guidelines of this project.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings.

---
*Note: Only Maintainers can approve and merge PRs. Expected initial review time: 3 business days.*


dataweaver/apps/web/src/components/primitives/icons/status_indicator.tsx
```diff
@@ -0,0 +1,12 @@
+import type { ComponentPropsWithRef } from 'react';
+import { mergeClassNames } from '~/functions/merge_class_names';
+import s from './status_indicator.module.scss';
+
+type StatusIndicatorProps = ComponentPropsWithRef<'div'>;
```
nick-nlb: Props should match component name `IconStatusIndicatorProps`, and prefer `interface` over `type` where possible (see conventions in `FRONTEND.md`.

dataweaver/apps/web/src/components/primitives/icons/status_indicator.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export const IconStatusIndicator = ({
+  className,
+  ...rest
```
nick-nlb, line 9: Other icons set `aria-hidden="true"` directly within the icon, whereas in this case, it is set to be true as a prop here. Is this intentional? If so, can we comment why (or otherwise, let's update this to match existing icons where it is set within).

dataweaver/apps/web/src/components/elements/button.module.scss
```diff
@@ -204,4 +211,12 @@
 
     padding-block: 10px;
   }
+
+  [data-size="huge"] & {
+    @include type-label-large;
+
+    padding-block: 16px;
+    font-size: 20px;
```
nick-nlb: This is a hard-coded override of the font-size and line-height. Can we move this to a mixin, following the pattern of the other sizes?

dataweaver/apps/web/src/components/elements/button.module.scss
```diff
@@ -102,12 +102,19 @@
...
-    &[data-size="large"] {
+    &[data-size="large"],
+    &[data-size="huge"] {
       column-gap: 6px;
       min-height: var(--button-pill-size-large);
       padding-inline: 18px;
       border-radius: calc(var(--button-pill-size-large) / 2);
     }
+
+    // "huge" shares large's padding/gap above but wants a fully pill-shaped
```
nick-nlb: Will this always be the case? I.e., as a matter of convenience, huge might currently share some attributes with large, but is that definitional for huge? 

Otherwise, this could be separated into two, with coincidentally shared features entered separately.

dataweaver/apps/web/src/components/scopes/page_home/intro.module.scss
```diff
@@ -1,10 +1,7 @@
 .outer-container {
-  display: grid;
-  overflow: clip;
   pointer-events: auto;
-  background: rgb(var(--color-query-surface-raised));
-  border-radius: var(--border-radius-large) var(--border-radius-medium);
-  box-shadow: var(--shadow-elevated);
+
+  @include pulsing-card(54px, false);
```
nick-nlb: Can we convert this from a magic number?

nick-nlb (commented):
Thank you for this update Pablo! Looks great. Some comments below.

dataweaver/apps/web/src/components/scopes/page_home/prompt.module.scss
```diff
@@ -87,7 +89,7 @@
   }
 
   &::placeholder {
-    color: rgb(var(--color-query-content-dimmed));
+    color: rgb(var(--color-query-content-subtle));
   }
```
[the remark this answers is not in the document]
nick-nlb: To help with immediate understanding of the history of the PR for reviewers, before resolving a comment (by either the auto-reviewer or a human reviewer), can we reply to the initial comment with an explaining of what was done, or if the suggestion was rejected, an explanation.

If it is simple (i.e., a simple suggestion that was followed and completed) an emoji would do as well.

dataweaver/.next/trace
```diff
@@ -0,0 +1 @@
+[{"name":"generate-buildid","duration":268,"timestamp":82282366368,"id":4,"parentId":1,"tags":{},"startTime":1787087319015,"traceId":"c5697aaaa0ba516f"},{"name":"load-custom-routes","duration":466,"timestamp":82282366761,"id":5,"parentId":1,"tags":{},"startTime":1787087319015,"traceId":"c5697aaaa0ba516f"},{"name":"create-dist-dir","duration":387,"timestamp":82282367254,"id":6,"parentId":1,"tags":{},"startTime":1787087319016,"traceId":"c5697aaaa0ba516f"},{"name":"clean","duration":312,"timestamp":82282368559,"id":7,"parentId":1,"tags":{},"startTime":1787087319017,"traceId":"c5697aaaa0ba516f"},{"name":"next-build","duration":900010,"timestamp":82281469012,"id":1,"tags":{"buildMode":"default","version":"16.3.1","bundler":"turbopack","failed":true},"startTime":1787087318118,"traceId":"c5697aaaa0ba516f"}]
```
nick-nlb: This file should be removed (it may be the result of running a build in the root of the project).

dataweaver/.next/trace-build
```diff
@@ -0,0 +1 @@
+[{"name":"next-build","duration":900010,"timestamp":82281469012,"id":1,"tags":{"buildMode":"default","version":"16.3.1","bundler":"turbopack","failed":true},"startTime":1787087318118,"traceId":"c5697aaaa0ba516f"}]
```
nick-nlb: This file also should be removed.

dataweaver/apps/web/src/components/elements/button.module.scss
```diff
@@ -102,17 +102,19 @@
...
       border-radius: calc(var(--button-pill-size-large) / 2);
     }
 
-    // "huge" shares large's padding/gap above but wants a fully pill-shaped
-    // (not just large-radius) corner, so override just the radius here.
+    // "huge" reuses large's height/padding but always renders as a full pill,
+    // so its radius isn't tied to (and shouldn't scale with) that height.
     &[data-size="huge"] {
+      column-gap: 6px;
+      min-height: var(--button-pill-size-large);
```
nick-nlb: This design token is still the one taken from the large button. Even if it happens to be the same, they represent semantically different values and should be separated.

Does this min-height have any effect in practice as it is? If the height of the button is always larger that the large min-height? (in practice this wouldn't matter if we move to a token for extra-large with the correct min-height).

dataweaver/apps/web/src/components/elements/button.tsx
```diff
@@ -27,7 +27,7 @@ interface WithIconOnly {
 
 interface WithChildrenAndOptionalIcon {
   children: ReactNode;
-  size: 'small' | 'medium' | 'large';
+  size: 'small' | 'medium' | 'large' | 'huge';
```
nick-nlb: We have a `size: 'small' | 'medium' | 'large' | 'extra-large';` size set for `WithIconOnly`, and have a `huge` size added in this PR for `WithChildrenAndOptionalIcon`.

These should be harmonized (I would go with extra-large over huge, as it simplifies naming for larger sizes).

dataweaver/apps/web/src/components/elements/button.module.scss
```diff
@@ -108,6 +108,15 @@
...
       border-radius: calc(var(--button-pill-size-large) / 2);
     }
+
+    // "huge" reuses large's height/padding but always renders as a full pill,
+    // so its radius isn't tied to (and shouldn't scale with) that height.
+    &[data-size="huge"] {
+      column-gap: 6px;
+      min-height: var(--button-pill-size-large);
+      padding-inline: 18px;
+      border-radius: 9999px;
```
nick-nlb: Now that we will have a dedicated pill size for huge (soon to be extra-large), do we need this difference between extra-large and the other sizes? 

Related to that, could we actually change all the buttons that are intended to be pills to have a border-radius of 9999px? (and move that to a token to make it less of a "magic number" - although 9999px is standard for pills).

So to sum up, we would:
* change all pill border radiuses to 9999px (with a token)
* separately, give huge (to be renamed extra-large) its own height token, and that height token should represent the actual min-height of the extra large button (which is larger than large).

dataweaver/apps/web/src/styles/includes/_typography.module.scss
```diff
@@ -114,6 +114,16 @@ $-supported-body-font-weights: 400, 500;
   }
 }
 
+@mixin type-label-huge($has-text-box-trim: true) {
+  font-size: 20px;
+  font-weight: 500;
+  line-height: 22px;
```
nick-nlb: Line height elsewhere in this file is unitless, whereas here this is in pixels. Is there a reason for this divergence?

An additional divergence is that this doesn't contain a breakpoint in size for mobile.

nick-nlb (commented):
Thank you for the updates! 

Some comments attached to this review.

dataweaver/apps/web/src/components/elements/button.module.scss
pablonoel: Agreed, changed it also, to extra-large, for convention

dataweaver/apps/web/src/styles/includes/_typography.module.scss
pablonoel: Changed to unitless, added the breakpoint

dataweaver/apps/web/src/components/elements/button.module.scss
```diff
@@ -204,4 +213,10 @@
 
     padding-block: 10px;
   }
+
+  [data-size="extra-large"] & {
+    @include type-label-huge;
```
nick-nlb: The migration away from huge is incomplete.

dataweaver/apps/web/src/styles/includes/_typography.module.scss
```diff
@@ -114,6 +114,20 @@ $-supported-body-font-weights: 400, 500;
   }
 }
 
+@mixin type-label-huge($has-text-box-trim: true) {
+  font-size: 20px;
```
nick-nlb: Is the font-sizing distribution between breakpoints intentional here?

Unlike other sizes, here the size is inverted, with larger screen sizes being 18px and smaller being 20px. This should be other other way around? (i.e. what is in the breakpoint include is what is larger than the breakpoint).

dataweaver/apps/web/src/styles/core/_root.scss
```diff
@@ -8,6 +8,7 @@
     --button-pill-size-small: 27px;
     --button-pill-size-medium: 32px;
     --button-pill-size-large: 34px;
+    --button-pill-size-extra-large: 34px;
```
nick-nlb: This doesn't represent the actual size of the button (which at its maximum size with the larger font, is 47px - although that can be revisited as you verify the font sizes re: the previous comment.

dataweaver/apps/web/src/components/scopes/page_home/prompt.module.scss
```diff
@@ -61,21 +61,27 @@
...
   --max-rows: 4;
 
-  @include type-body-large;
+  @include type-body-large(500);
 
   width: 100%;
   max-height: calc(var(--max-rows) * 1lh);
   overflow-y: auto;
-  color: rgb(var(--color-query-content));
+  font-size: 20px;
```
nick-nlb: Isn't this redundant with the inclusion of `type-body-large` above?

dataweaver/apps/web/src/components/elements/button.module.scss
```diff
@@ -108,6 +108,15 @@
       padding-inline: 18px;
       border-radius: calc(var(--button-pill-size-large) / 2);
```
nick-nlb: Can we apply the "border-radius-full" token to all buttons?

dataweaver/apps/web/src/styles/core/_root.scss
nick-nlb: It looks like this issue is unresolved.

dataweaver/apps/web/src/components/elements/button.module.scss
```diff
@@ -92,21 +92,30 @@
...
     &[data-size="large"] {
       column-gap: 6px;
       min-height: var(--button-pill-size-large);
       padding-inline: 18px;
-      border-radius: calc(var(--button-pill-size-large) / 2);
+      border-radius: var(--border-radius-full);
+    }
+
+    // "extra-large" reuses large's height/padding but always renders as a full pill,
+    // so its radius isn't tied to (and shouldn't scale with) that height.
```
nick-nlb: This comment is now out of date (given that all the buttons follow the same principle for the radius, and that this size now has its own height).

nick-nlb (approved):
Thank you Pablo! There's still one (minor) comment to address (the now out-of-date comment), and then LGTM

PR #444, opened 2026-08-17
Narratives services image

ddebasmita-lab:
## Overview

Clear description of the changes.

Dockerfile — a thin overlay on the upstream Custom Data Commons image. Adds the compiled React app at /workspace/react-ui/ and replaces the nginx config; upstream's entrypoint, which supervises nginx, Flask, Mixer, MCP and the NL server, is left untouched.

The React build runs in a node stage inside the image rather than on the host. That keeps a clean clone reproducible — npm ci installs exactly the tree in package-lock.json — and avoids a pre-built dist/ having to be staged first. It also sidesteps .dockerignore, which excludes dist/ and would silently break a host-staged copy.
nginx.conf — routes:
| Path            | Goes to                             |
|-----------------|-------------------------------------|
| /agent/*        | the sidecar on [127.0.0.1:5001](http://127.0.0.1:5001/)       |
| /, /assets/*    | the React build, from the image     |
| /core/api/*     | Mixer on :8081, with CORS preflight |
| /mcp            | the MCP server on :8082             |
| everything else | upstream Flask on :7070             |

### DIFF
--- a/narratives/.dockerignore
+++ b/narratives/.dockerignore
@@ -1,6 +1,13 @@
 node_modules
 dist
 .git
+
+# The services image builds only the React app. These sit inside the build
+# context but contribute nothing to it. Patterns match the context root only,
+# so src/config/ is unaffected.
+agent
+config
+
 .env
 .env.local
 .env.*.local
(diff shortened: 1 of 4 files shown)


juliawu (approved):
Nice!

PR #447, opened 2026-08-22
feat(data-weaver): child place resolution

nick-nlb:
## Issue

This PR addresses tracker lines 10, 11, 16 and 19.

## Description

The goal of this PR is to update place resolution across the query execution pipeline, store and front-end in order to support sub-region and child place retrieval and multi-series chart visualizations. 

An important note is that while this PR does update skills and agent directions in order to improve the results of searches involvin child places, its primary goal is to create the update data fetching and the end-to-end pipeline to support child searching and display. We will continue to improve the skills in future PRs.

## Implementation

* Updated agent instructions in order to have the agent able to understand whether a query should be a child place query.
* Fixed store structure and query pipeline so that the connection between the actual place resolved and the stat var is preserved (this addresses the otherwise very frequent "no data" issues). (Addressing items 10 and 16)
* Implemented child place hierarchy retrieval using the Data Commons /v2/node API (<-containedInPlace+) to fetch all sub-entities for a given parent region. As part of this, introduced place normalization to Data Commons place types for reliable child fetching.
* Extended time-series observation fetching to support multi-entity requests across child places.
* Implemented composite scope key generation to keep canvas cards, history nodes, and the store synchronized.
* Centralized card title formatting and removed fragile title parsing. (Addressing item 19)

## Notes

Upcoming items, to be implemented in separate PRs:

* Implementation of choropleth and ranking tiles (both better suited to handle large numbers of child places).
* Limits to the number of series that display on a line chart (e.g. "Displaying top x…").
* General improvements to line charts and tooltips.
* Refactoring of query stream handling (right now, this function is very long, and longer still with this PR).

## Tests

Added unit tests for place type normalization, scope key derivation, and title formatting.

## Testing

"How access to electricity has changed across countries in Africa"

### On main branch

* In main, the agent does not recognize the "child resolution" nature of the query and either produces unsatisfactory results or, worse, results in "No Data" because of a mismatch between the entities and stat vars.
* The reason for the "No Data" is that in main, the model looks for data for the primary entity ("Africa"), does not find it, and falls back to "Earth" where it finds data (and therefore stat vars). However, it loses the connection between the stat var and the actual entity found, and attempts to apply it to Africa where there is no data. This core issue is fixed (along with the improvement to perform a child place search rather than fall back to Earth).

### In this PR

The search will resolve to child places, and display all countries on a line chart.
* Note: the line chart is not the best tool for this type of data: this will be addressed in upcoming PRs.
* The search itself, being deterministic, does not always perform a child search. We can harden this in subsequent PRs.


dataweaver/apps/web/src/server/steps/render_result_html.ts
```diff
@@ -48,11 +50,18 @@ const buildTableHtml = (result: QueryResult): string => {
         ? facets.map(formatFacetBlock).join('<br><br>').replace(/\|/g, '\\|')
         : 'No data';
 
+    const scopeKey = variable.isChildQuery
+      ? variable.parentPlaceDcid || variable.placeDcid
+        ? `${variable.parentPlaceDcid || variable.placeDcid}:${variable.childPlaceType || 'Country'}`
+        : defaultScopeKey
+      : variable.placeDcid || defaultScopeKey;
```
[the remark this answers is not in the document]
nick-nlb: Done as part of addressing the previous comment

dataweaver/apps/web/src/server/steps/observations.ts
```diff
@@ -99,36 +102,52 @@ export const fetchTimeSeriesBatch = async (
...
+        const facets: FacetInfo[] = orderedFacets.map((facet) => {
+          const facetData = facetsMap[facet.facetId] || {};
+          const metadata = metadataMap[facet.facetId];
+
+          // Data Commons API does not guarantee chronological ordering of the
+          // observations array within a facet. Sort ascending by date so downstream
+          // consumers (charts, comparisons, latest value lookups) have a guaranteed order.
+          const observations = (facet.observations || [])
+            .slice()
+            .sort((a, b) => a.date.localeCompare(b.date));
```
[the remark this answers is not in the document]
nick-nlb: Done.

dataweaver/apps/web/src/server/config/skills/data_discovery.md
```diff
@@ -7,8 +7,14 @@ maxToolCalls: 10
...
+     - World / Continent (e.g. "Earth", "Africa", "Europe", "Asia") -> infer `childPlaceType: "Country"`
+     - Nation / Country (e.g. "United States", "India", "Canada") -> infer `childPlaceType: "State"` (or "AdministrativeArea1")
+     - State / Province (e.g. "California", "Texas", "Ontario") -> infer `childPlaceType: "County"` (or "AdministrativeArea2")
+   - **Action for Child Queries**: Do NOT use `search_indicators` or `get_observations` for the parent alone. Call `search_child_indicators` (with `parent_place` DCID and `sample_child_places` to test if indicators exist), inspect metadata with `get_variable_metadata`, and fetch data across all children with `get_child_observations` (passing `parent_place_dcid` and `child_place_type`, e.g. `parent_place_dcid: "country/USA"`, `child_place_type: "County"`, `date: "all"`, or `parent_place_dcid: "africa"`, `child_place_type: "Country"`, `date: "all"`). Set `"isChildQuery": true`, `"parentPlaceDcid": "<parent_dcid>"`, and `"childPlaceType": "<type>"` on each variable in the `variables` array.
```
beets, line 17: i wonder if these instructions can be made into a bulleted list (easier for us and the LLM to parse).

Something to consider for follow ons.

dataweaver/apps/web/src/server/config/skills/data_discovery.md
```diff
@@ -48,10 +59,11 @@ JSON SCHEMA:
...
+   - **For Single-Place Queries**: Before adding any variable to `variables`, you MUST verify that the target place's DCID is present in `places_with_data` in the `search_indicators` tool output for that variable.
+   - **For Sub-Region / Child-Place Queries** (e.g. "counties in California", "California counties", "unemployment across counties in the USA", "cities across the US", "income across California", "countries in Africa", "states in USA"): An explicit child place type ALWAYS wins over any default hierarchy. Do NOT use `search_indicators` with the continent/parent alone. You MUST use `search_child_indicators` (passing `sample_child_places` like `["geoId/06001", "geoId/48001"]` for counties, `["geoId/06", "geoId/48"]` for states, or `["country/NGA", "country/ZAF"]` for countries) and `get_child_observations` (passing `parent_place_dcid` and `child_place_type`, e.g. `parent_place_dcid: "country/USA"`, `child_place_type: "County"`, `date: "all"`). Set `isChildQuery: true`, `parentPlaceDcid`, and `childPlaceType` on the returned variable.
```
beets, line 65: This seems duplicative with the instructions above. Have you observed regressions without this restatement?

dataweaver/apps/web/src/app/api/query/route.ts
beets, line 1: In a follow on, let's do a pure refactor on this file to pull out the processing for each stage into helpers.

dataweaver/apps/web/src/functions/normalize_place_type.ts
```diff
@@ -0,0 +1,155 @@
+/**
+ * Canonical mapping from common aliases, plurals, and variations
+ * to official Data Commons schema Place types (PascalCase).
+ * Sourced from Data Commons constants and place hierarchies.
+ */
+const PLACE_TYPE_MAP: Record<string, string> = {
```
beets, line 9: This will slowly get out of sync with our KG. Fine for now, but we should work on figuring out uplifting this logic to the MCP server / skill that we ship. Otherwise this would have to be done by all the MCP server clients...

beets (approved):
Thanks for this, and nice to see more tests! Left comments, but many of these aren't gating -- more for follow ons. Can be captured in TODOs.

dataweaver/apps/web/src/app/api/query/route.ts
nick-nlb: Absolutely! This is actually already on the tracker, but I will prioritize it for a fast-follow from the demo.

dataweaver/apps/web/src/functions/normalize_place_type.ts
nick-nlb: Agreed - this should be JSON, and moved upstream. Added to the tracker for once it has stabilized and been stress-tested in this repo.

dataweaver/apps/web/src/server/config/skills/data_discovery.md
nick-nlb: Adding this did improve the agent's behavior. Although the instructions are very similar (and in the same file), they instruct the agent at different stages of the loop.

This does highlight what we were discussing earlier, which is that there is a lot of room for experimentation (and potential to break up this discovery instruction set into multiple that get fed and read at different times to keep a cleaner context for the agent).

Once we have an eval framework, we could do a fair amount of experimentation here.

dataweaver/apps/web/src/server/config/skills/data_discovery.md
nick-nlb: Added this to the tracker as well for a fast-follow (although perhaps best for once we have evals in place).

PR #454, opened 2026-08-28
chore(data-weaver): add top-level orchestrator GitHub Actions CI workflow and associated Data Weaver workflow

nick-nlb:
## Description

This PR adds top-level CI orchestration workflow, and an associated workflow for Data Weaver. 

Data Weaver's workflow automatically validates pull requests and branch pushes with type checks, linting, unit tests, and production build checks.

The goal of this architecture is to allow individual applications within the monorepo to provide CI workflows with blocking tests in a way that does not affect other applications. Currently, only Data Weaver has such tests.


.github/workflows/ci.yml
beets, line 1: Please add top-level comments, especially describing the need for this architecture vs a single CI per app

beets (approved):

PR #458, opened 2026-09-01
feat(data-weaver): choropleth charts

nick-nlb:
## Description

This PR adds choropleth map visualization for multi-entity geographic comparisons in Data Weaver.

This fills the use case where data for a large number of places, usually children, needs to be displayed on a single chart. 

## Implementation
*   **Geographic Boundaries API & Service**: Implemented `/api/geo` route to query GeoJSON from the Data Commons API.
*  **LRU caching** (via a library): In the same session, we often use the same maps (i.e. a map of Africa with different data on different cards). To avoid refetching, we use a byte-limited LRU (via the very widely used lru-cache library).
*   **Map Projections & Bounds Fitting**: Added regional projection heuristics (`map_projection.ts`) supporting curated regional projections (Lambert Azimuthal for Europe, rotated Equirectangular for Asia/Oceania, Albers USA for US states) and dynamic bounds fitting (`map_bounds.ts`). This is largely based on the Data Commons website implementation (with some additional landmass filtering).
*   **Choropleth Component & Time Slider**: Created `DataChartChoropleth` with SVG pan/zoom, interactive tooltip with value/unit formatting, color ramp legend, and `SliderTime` for stepping or auto-playing through temporal observations.
*   **General Pipeline**: Updated pipeline and store to handle parentPlaceId and chropleths, added styling for the new components, updated prompts to recognize map-appropriate queries.

## Testing
*   **Build & Type Validation**: Verified zero errors across `pnpm test`, `pnpm lint` (TypeScript + Biome + Stylelint), and `pnpm build`.

## Coming Functionality (all in Tracker)

* Better height handling during loading (to avoid layout shift).
* Clickable spawning of line charts from choropleth
* A consolidated facet selector (i.e. no very tall cards like we see now).
* Investigate GeoJSON simplification (the calls are very quick and responsive, but the GeoJSON is often much more detailed than needs to be given the card size, with the caveat that we do need to accommodate zooming).
* More color schemes for the choropleths (currently a single monotonic choropleth color scheme available).


dataweaver/apps/web/src/components/elements/card/chart/data_chart_choropleth.tsx
```diff
@@ -0,0 +1,639 @@
...
+      role="img"
+      aria-label="Choropleth map"
+      className={s['map-svg']}
+      viewBox={`0 0 ${width} ${height}`}
+      onWheel={onWheel}
+      onPointerDown={onPointerDown}
+      onPointerMove={onPointerMove}
+      onPointerUp={onPointerUp}
+      onPointerLeave={onPointerLeave}
+    >
```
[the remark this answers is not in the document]
nick-nlb: Agreed (although I don't like the lose typing in the solution above). Done.

dataweaver/apps/web/src/app/api/geo/route.ts
```diff
@@ -0,0 +1,260 @@
...
+          for (const n of childNodes) {
+            if (
+              n.dcid &&
+              !featureMap.has(n.dcid) &&
+              (n.types?.includes('Country') ||
+                n.types?.includes('State') ||
+                n.types?.includes('County') ||
+                n.types?.includes('Place') ||
+                n.types?.includes('AdministrativeArea1') ||
+                n.types?.includes('AdministrativeArea2'))
```
beets, line 172: let's pull these out as constants

would really prefer if this logic did not live in the app (but in our API / client library), so could you add a TODO

beets: similarly, logic for pulling down the right resolution level, adding the unwinding logic at processing time, etc.

dataweaver/apps/web/src/components/elements/card/chart/chart.tsx
```diff
@@ -155,12 +165,69 @@ export const CardChart = ({
...
+  const validEntityKeys = useMemo(
+    () =>
+      chartSeries
+        ?.map((s) => s.key)
+        .filter((k) => k && k !== 'default')
+        .sort() ?? [],
+    [chartSeries],
+  );
+
+  const [isGeoAvailable, setIsGeoAvailable] = useState<boolean | null>(() => {
```
beets, line 177: should we pull out the choropleth logic into a subclass (as static helpers perhaps).

dataweaver/apps/web/src/components/elements/card/chart/data_chart_choropleth.tsx
beets, line 1: please add documentation to the module

dataweaver/apps/web/src/components/elements/card/chart/data_chart_choropleth.tsx
```diff
@@ -0,0 +1,651 @@
...
+        )}
+
+        {hovered && !isDragging && (
+          <div
+            className={[
+              s.tooltip,
+              hovered.y < 60 ? s['tooltip-align-bottom'] : '',
+              hovered.x < 80 ? s['tooltip-align-left'] : '',
+              hovered.containerWidth !== undefined &&
+              hovered.x > hovered.containerWidth - 80
```
beets, line 598: please pull out these values as constants. best is if they could be computed.

beets (approved):
thanks for this! ideally, a lot of this logic needn't live in dataweaver but is something that is easily plugged in. so i'd keep that in mind as we add this here first, that we'd lift this out of the app soon.

dataweaver/apps/web/src/app/api/geo/route.ts
nick-nlb: Done! (pulled out as constants and added the todo - we can brainstorm in the long run where this type of logic should go, as we can either move it up (to the SDK, but that means client side) or down (to the APIs).

dataweaver/apps/web/src/components/elements/card/chart/data_chart_choropleth.tsx
nick-nlb: The comment that these should be computed had be revisit the tooltips, and do a refactor of them. Not only are these now computed, but we no longer have two separate charting tooltips (for choropleths and all the others).

dataweaver/apps/web/src/components/elements/card/chart/chart.tsx
nick-nlb: I've pulled out the logic into a hook for the choropleth logic and a set of static functions in their own file.

PR #459, opened 2026-09-02
feat(data-weaver): add friendly tool call labels and remove misleading counts

nick-nlb:
## Description

This PR maps low-level Data Commons MCP tool identifiers to human-readable status labels during streaming queries to improve the user-facing discovery experience.

It also removes the display of the model call iteration count and the denominator (count/max), because the max was misleading, as it was the maximum allowable calls, not the expected number (and the searches usually resolve in far fewer calls than the maximum).

## Notes

The friendly labels for tool calls were chosen with the intention of providing clean looking descriptions that would make sense to the user. They are open to other options. 

## Testing
* Added unit tests in `apps/web/src/server/status.test.ts` covering known tool mappings and fallback behavior.
* Verified production build and production build runtime via `pnpm build` and `pnpm preview`.


beets (approved):
🙌

